PR #2, opened 2021-05-05
add all packages defined in overlays to devshell

Pacman99:
All overlays passed to the default channel will be auto-added to the devshell. This gives users a means to customize their devshell.


blaggacao:
I'm not sure, what our model for user facing devshell modifications would be. But it seems like we want to re-use the lib devshell in a way? In such scenario this makes sense.

Pacman99:
I think this is a good first step to allow for customizing devshell. But it would be nice to add something to the API to provide even better customizations. Since devshell also uses module system, we could even do it as a submodule and just add it to `imports` line of the call to devshell.

`integrations.devshell` or something similar.

Pacman99:
I switched this to actually use the overlays exporter too. So it follows the exact logic we use to export packages. Except overlays from `inputs` aren't filtered out.

Pacman99:
And I think we should re-use the lib devshell. Any thing we use for development will likely be useful for users too.

blaggacao:
Ah ok, got it.

PR #4, opened 2021-05-12
imp: remove importer from top level lib exports

blaggacao:
Importers are auxiliary functions of this library. It feels inconsistent when
they have the same visibility of primary (builder) functions.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -50,8 +50,6 @@
         inherit attrs lists modules importers generators;
         inherit (lib)
           mkFlake
-          pathsIn
-          importHosts
           mkDeployNodes
           mkHomeConfigurations;
       };


Pacman99:
I don't understand this, why are they any less important than builder functions? And why should they get less visibility?

blaggacao:
> I don't understand this, why are they any less important than builder functions? And why should they get less visibility?

Isn't that what `importers` is for? Everything has its well-defined place.

Under the `importers` namespace, it is clear to everybody to what category they belong to (and what they do). It is more confusing when they are mixed with generators on the top level namespace.

I think though, generators should reside on the top level.

Pacman99:
> I think though, generators should reside on the top level.

This makes sense, I was confused as to how we would decide top-level exports in the future. But with this defined convention, I agree with this PR.

PR #5, opened 2021-05-12
Rearrange, cleanup, and improve docs

Pacman99:
Moved all nix code to the src/ folder. And added mkFlake options documentation in the doc/ folder. We can make use of this folder in the future for other things too.

Also remove unnecessary devos and compat folders.


blaggacao (approved):
Excellent!

PR #6, opened 2021-05-12
test mkFlake with a full flake similar to devos

Pacman99:
Starting point for tests, so you can mostly trust a `nix flake check` success. To test more things, we can add more "checks" in fullFlake/default.nix. So you could test if `ourlib` works right, or see if other outputs are evaluated right in mkFlake/evalArgs by forcing those things to be evaluated in a module or overlay. But this PR is just to get that framework setup.


Pacman99:
bors try

Pacman99:
bors try

Pacman99:
bors try

Pacman99:
bors try

blaggacao:
Would it make sense to make the `fullFlake` a wrapped git submodules of `divnix/devos`?

That way we can relatively easily ensure that any given version of `devlib` does not break any given version of `devos`.

Pacman99:
> Would it make sense to make the fullFlake a wrapped git submodules of divnix/devos?

That was my first idea for this! I chose not to for a couple reasons. It reduces a lot of customization with testing, I imagine we can keep adding more checks for specific portions of mkFlake by editing the default.nix. We could export specific packages dedicated for checking things. Also I'm, probably irrationally, a little worried about circular dependencies.

Pacman99:
Also we would have to incur more depencies on devlib, since devos depends on home-manager and other things.

blaggacao:
Alright. Yeah, that was just a rather random idea. Makes absolute sense not to introduce additional dependencies here.

If you feel that this is ready, please go ahead.

bors try

Pacman99:
I want to get bors to merge this PR, pretty much the only blocker to me

nrdxp:
bors try

nrdxp (approved):
LGTM

PR #8, opened 2021-05-14
tree: rename devlib -> digga

Pacman99:


src/modules.nix
```diff
@@ -45,7 +45,7 @@
 
       _module.args = {
         inherit self;
-        devlib = pkgs.devlib;
+        diggaLib = pkgs.lib.digga;
```
blaggacao, line 48: `# diggaLib can be accessed in modules directly as diggaLib`

blaggacao (approved):
I liked the friendly comment on the other site so much, I'd like to see it here as well. :smile:

src/modules.nix
Pacman99: sounds good

Pacman99: changed it to:
`# digga lib can be accessed in modules directly as diggaLib`
thats more accurate

PR #11, opened 2021-05-15
add devshellModules option to customize devshell

Pacman99:
Adds `devshellModules` api argument to add modules to the devshell. This also accepts `.toml` files which get auto-imported with devshell's `devshell.importToml`. To get that to work we have to pass `pkgs` as a specialArg so it can be used in an `imports` line.

### DIFF
--- a/src/mkFlake/default.nix
+++ b/src/mkFlake/default.nix
@@ -96,6 +96,7 @@ lib.systemFlake (lib.mergeAny
     devShellBuilder = channels:
       lib.pkgs-lib.shell {
         pkgs = getDefaultChannel channels;
+        extraModules = cfg.devshellModules;
       };
   }
   otherArguments # for overlays list order
--- a/src/mkFlake/evalArgs.nix
+++ b/src/mkFlake/evalArgs.nix
@@ -43,6 +43,16 @@ let
       coercedListOf = elemType: with types;
         coercedTo anything (x: flatten (singleton x)) (listOf elemType);
 
+      pathToDevshellModule = with types; (coercedTo path (path:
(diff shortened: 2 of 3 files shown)


src/importers.nix
```diff
@@ -49,9 +49,19 @@ let
     in
     map fullPath (lib.attrNames (lib.safeReadDir dir));
 
+  devshellOverlay = module:
+    {
+      # Meant to output a module that sets the overlays option
+      overlays = [
+        (final: prev: {
+          devshellAdditions = prev.devshellAdditions or [ ]
+            ++ [ module ];
```
Pacman99, line 58: Annoying part about doing this is as part of overlays is that lists get overriden.I think this is a good way to get around that.

src/importers.nix
```diff
@@ -49,9 +49,19 @@ let
     in
     map fullPath (lib.attrNames (lib.safeReadDir dir));
 
+  devshellOverlay = module:
+    {
+      # Meant to output a module that sets the overlays option
+      overlays = [
+        (final: prev: {
+          devshellAdditions = prev.devshellAdditions or [ ]
```
blaggacao, line 57: `___dontExport = true;` ?

blaggacao (commented):
I guess, we want to avoid to also export this (implementation detail).

src/importers.nix
blaggacao: `__devshellAdditions` ?

To make an accidental name clash even more unlikely by orders of magnitude through the magic of an internal property marker (`__`)?

blaggacao (approved):
Exzellent!

PR #14, opened 2021-05-18
Revert "shell: add packages from overlays to devshell"

Pacman99:
This reverts commit be058ad58dd0e43c46138a073e94ca3c313401be.

In hindsight, this is a bit too eager especially for those who pull larger overlays. The changes from #11 provide a much better method to customize your devshell. I think for now we can just leave it at you can't customize the devshell. The feature will be added in the next release with #11.

### DIFF
--- a/src/pkgs-lib/shell/default.nix
+++ b/src/pkgs-lib/shell/default.nix
@@ -17,7 +17,10 @@ let
     })
   ];
 
-  pkgs' = pkgs.appendOverlays overlays;
+  pkgs' = import pkgs.path {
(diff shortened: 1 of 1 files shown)


Pacman99:
Referring to this comment: https://github.com/divnix/devos/issues/283#issuecomment-843246939

@benneti pulled the entire emacs overlay. And it makes no sense to build that whole thing. Large overlays are common enough to show that this is a bad design, but its fine we have a much better one now!

Pacman99:
bors try

blaggacao (approved):

PR #15, opened 2021-05-19
mkFlake: update to new fup outputsBuilder

Pacman99:
following changes from here: https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/59

### DIFF
--- a/src/mkFlake/default.nix
+++ b/src/mkFlake/default.nix
@@ -83,21 +83,29 @@ lib.systemFlake (lib.mergeAny
       inherit (self) pkgs;
     };
 
-    packagesBuilder = lib.exporters.fromOverlays self.overlays;
+    outputsBuilder = channels:
+      let
+        defaultChannel = channels.${cfg.nixos.hostDefaults.channelName};
+      in
+      lib.mergeAny
+        {
+          packages = lib.exporters.fromOverlays self.overlays channels;
 
-    checksBuilder = channels:
-      lib.pkgs-lib.tests.mkChecks {
-        pkgs = getDefaultChannel channels;
-        inherit (self.deploy) nodes;
-        hosts = self.nixosConfigurations;
-        homes = self.homeConfigurations or { };
-      };
+          checks = lib.pkgs-lib.tests.mkChecks {
+            pkgs = defaultChannel;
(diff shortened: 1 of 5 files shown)


Pacman99:
Also adds the feature that digga users can write their own checks or do their own logic for package exporting, which will get merged with what digga does by default.

But this would make it impossible to remove checks.

blaggacao (approved):
LGTM. Would it make sense at this occasion to add checks foe the different categories of other arguments that sould get merged? Just to make sure we don't break anything unadvertedly in the future.

src/mkFlake/default.nix
```diff
@@ -83,21 +83,29 @@ lib.systemFlake (lib.mergeAny
...
+            hosts = self.nixosConfigurations;
+            homes = self.homeConfigurations or { };
+          };
+
+          devShell = lib.pkgs-lib.shell {
+            pkgs = defaultChannel;
+            extraModules = cfg.devshellModules;
+          };
+        }
+        (otherArguments.outputsBuilder channels);
```
Pacman99, line 106: Would it be ok to add `outputsBuilder` to evalArgs? Otherwise if someone passes a `outputsBuilder` it would actuallly override digga's outputsBuilder. Since `otherArguments` is on the right side of the merge. Or we can just move otherArguments to the left side.

Pacman99: I would prefer adding `outputsBuilder` as an official argument in `evalArgs`.

Pacman99:
> Would it make sense at this occasion to add checks foe the different categories of other arguments that sould get merged? Just to make sure we don't break anything unadvertedly in the future.

Definitely would, I'll write up something for that soon.

src/mkFlake/default.nix
Pacman99:
```suggestion
        (cfg.outputsBuilder channels);
```

blaggacao: > I would prefer adding outputsBuilder as an official argument in evalArgs.

That sounds good to me :+1:

Pacman99:
bors try

blaggacao (approved):

PR #20, opened 2021-05-21
Improve the secrets experience with devshell

GTrunSec:
This PR is related to https://github.com/divnix/devos/pull/279. It may help users to easily use the agenix and sops-nix.

Before that, We should know what is different between agenix and sops-nix.

- https://nixos.wiki/wiki/Comparison_of_secret_managing_schemes
- https://discourse.nixos.org/t/comparison-of-different-key-secret-managing-schemes/12001

### DIFF
--- a/src/pkgs-lib/shell/devshell.toml
+++ b/src/pkgs-lib/shell/devshell.toml
@@ -25,5 +25,18 @@ category = "documentation"
 package = "mdbook"
 category = "documentation"
 
+[[commands]]
+name = "ssh-show-ed25519"
+category = "secrets"
+command = '''
+          user="$(cut -d ';' -f 1 <<< "$1")"
+          if [[ "$user" == "root" ]]; then
+          ssh $1 "cat /etc/ssh/ssh_host_ed25519_key.pub"
+          else
+          ssh $1 "cat ~/.ssh/id_ed25519.pub"
+          fi
+        '''
+help = "ssh-show-ed25519 <user@hostName> | Show target host's SSH ed25519 key"
+
 [git.hooks]
 enable = true


blaggacao:
Nice! Thank you! We can implement these kind of solutions either in `digga` or in the `devos` template.

I'd be reluctant to choose any particular implementation here in `digga`, but if we were to choose, I'd probably prefer `agenix` for two reasons:

- It was specifically built to use `age` ("actually good encryption").
- It was specifically built to work with nix files, natively.

A little more context: https://github.com/divnix/devos/discussions/185#discussioncomment-486731

---

Now, to implement devshell additions in the "user space" of the template, we have https://github.com/divnix/digga/pull/11 which introduces a top level argument to `mkFlake`, called `devshellModules`, which will be amended to the devshell. For the moment, it's only on digga's develop branch.

We should probably implement some of the utilities you are proposing using `devshellModules` in https://github.com/divnix/devos/pull/279 (for agenix).

@GTrunSec What do you think about my thoughts, here?

src/pkgs-lib/shell/devshell.toml
```diff
@@ -25,5 +26,46 @@ category = "documentation"
...
+category = "secrets"
+command = '''
+          user="$(cut -d ';' -f 1 <<< "$1")"
+          if [[ "$user" == "root" ]]; then
+          ssh $1 "cat /etc/ssh/ssh_host_ed25519_key.pub"
+          else
+          ssh $1 "cat ~/.ssh/id_ed25519.pub"
+          fi
+        '''
+help = "ssh-show-ed25519 <user@hostName> | Show target host's SSH ed25519 key"
```
Pacman99, line 40: I think this is the only command I currently support adding here - maybe you can change my mind.

For the rest of it, I see their usefulness. But I don't think we should add it by default for the sake of keeping digga more general. And especially now that its possible to customize the devshell with #11, I think it makes sense to just use that for sops functions since neither devos nor digga supports sops.

GTrunSec:
>I'd be reluctant to choose any particular implementation here in digga, but if we were to choose, I'd probably prefer agenix for two reasons:

Agree with that, The agenix would be a better choice as the default solution of secret. Now I realize that adding sops-nix is ​​unnecessary at this time;

>In future we will also support cloud key management APIs such as AWS KMS, GCP KMS, Azure Key Vault or Hashicorp's vault.

Maybe one day the sops-nix achieved some features as following, that will be a good time point to use it as the consideration of the second set of plans;

>We should probably implement some of the utilities you are proposing using devshellModules as follow-up to divnix/devos#279 (for agenix).

looks like it is. The `devshellModules` seems more suitable for my current situation. Also, Adding sops-nix support is allowed more users to have another choice to use and know it, which would speed up the schedule of development for the new features.

src/pkgs-lib/shell/devshell.toml
GTrunSec: >For the rest of it, I see their usefulness. But I don't think we should add it by default for the sake of keeping digga more general.

Agree, I'm going to re-force push it. thanks for your suggestion.

Pacman99 (approved):
LGTM

blaggacao:
bors r+

Nice addition!

PR #24, opened 2021-05-21
Shell: setup export of devshell modules

Pacman99:
Introduces `devshell` namespace for `modules` and `externalModules`.

the import of .toml files no longer happens in evalArgs but instead happens as devshell modules are being exported and imported into shell.

depends on numtide/devshell#117

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -23,11 +23,11 @@
     },
     "devshell": {
       "locked": {
-        "lastModified": 1618523768,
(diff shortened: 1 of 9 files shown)


blaggacao:
This also satisfies the need behind https://github.com/numtide/devshell/issues/52

Pacman99:
bors try

Pacman99:
Also I'm not planning to add a deprecation message for `devshellModules` since the option was only ever added in `develop`.

blaggacao (approved):
Nice!

PR #29, opened 2021-05-23
rakeLeaves: create and use standard auto-importer

### LINKED ISSUE #25: NixOS Veteran Migration
We might consider to allow:

- `./hosts/hostname/configuration.nix`
- `./hosts/hostname/default.nix`

When NixOS Veteran "Jeff" wants to migrate, they can get a devos based system identical to their current setup by just moving the contents of `/etc/nixos` into `./hosts/hostname/` and then start factoring slowly from there, one piece at a time.

@Radvendii

### LINKED ISSUE #26: Overlays / Imports and treatment of non-nix files
People want to place a `.patch` used in overlays in the `./overlays` subfolder.

The current (optional) magic importer does not handle this case (and other non-nix file cases) gracefully.

@Radvendii

Pacman99:
The function can support files with .nix extensions along with folders. This allows it to be used for collecting overlays, hosts, and profiles

fixes #26
fixes #25

The output format of `rakeLeaves` is a combination of the old `mkProfileAttrs` and `recImport`.
Example file structure:
```
./core/default.nix
./base.nix
./main/dev.nix
./main/os/default.nix
```
Example output:
```nix
{
  core = ./core;
  base = ./base.nix;
  main = {
    dev = ./main/dev.nix;
    os = ./main/os;
  };
}
```


nrdxp:
Attempting to use this, I get this when importing the flake n a repl:
```console
error: infinite recursion encountered

       at /nix/store/gzl0zs2n25gidj5xqkch4svxvz0vkflf-source/lib/modules.nix:305:28:

          304|         builtins.addErrorContext (context name)
          305|           (args.${name} or config._module.args.${name})
             |                            ^
          306|       ) (lib.functionArgs f);
```

nrdxp (approved):
I am currently using this successfully to break up my host into [multiple files](https://github.com/nrdxp/devos/tree/nrd/hosts/serval-ws).
LGTM

src/importers.nix
```diff
@@ -1,73 +1,77 @@
...
+      ```
+      ./core/default.nix
+      ./core/dev.nix
+      ./base.nix
+      ```
+      Example output:
+      ```
+      {
+      core = {
+      default = ./core/default.nix;
```
blaggacao, line 25: I think in the context of `./profiles`, this requirement (to place a `default.nix` in a subfolder) is not really sufferable.

blaggacao (commented):
I think we need to defer to  https://github.com/divnix/digga/discussions/28#discussioncomment-779080 (et al.) for a round of discussions on the use cases.

src/importers.nix
Pacman99: `./profiles` doesn't need a `default.nix` only further directories. The goal here is contract validation, every node of the tree should map to a proper path. The best way to do that is to require that `.default` is a path or doesn't exist.

Pacman99:
Update: `mkProfileAttrs` now stops recursing when it hits a folder with a `default.nix`. So now there are nodes(contain leaves) or leaves(paths). A part of the tree cannot contain both.
Other changes:
 - moved `profileMap` to `importers.nix` its so related to mkProfileAttrs
 - we throw an error if non-strings are passed to suites, or if there are folders without `default.nix` in overlays or hosts. Maybe instead of this we can just fallback to "empty" overlays or hosts. So `_: _: { }` or `{ }` so you can just create random folders for other reasons in the two directories.

Pacman99:
bors try

src/importers.nix
```diff
@@ -1,73 +1,84 @@
...
+
+  profileMap = list: map
+    (p: getProfilePath (throw "passed profile ${builtins.toJSON p} isn't a path") p)
+    (lib.flatten list);
 
   overlays = dir:
     {
       # Meant to output a module that sets the overlays option
-      overlays = pathsIn dir;
+      # Only get top-level .nix files or default.nix from directories
```
blaggacao, line 69: I guess those comments are now inaccurate?

src/importers.nix
```diff
@@ -1,73 +1,84 @@
...
       # Meant to output a module that sets the hosts option
-      hosts = recImport {
-        inherit dir;
-        _import = base: {
-          modules = import "${toString dir}/${base}.nix";
-        };
-      };
+      hosts = lib.mapAttrs
+        (n: p: {
+          # Only get top-level .nix files or default.nix from directories
```
blaggacao, line 78: Same

src/importers.nix
```diff
@@ -1,73 +1,84 @@
...
+  profileMap = list: map
+    (p: getProfilePath (throw "passed profile ${builtins.toJSON p} isn't a path") p)
+    (lib.flatten list);
 
   overlays = dir:
     {
       # Meant to output a module that sets the overlays option
-      overlays = pathsIn dir;
+      # Only get top-level .nix files or default.nix from directories
+      overlays = map (getProfilePath (_: _: { })) (builtins.attrValues (mkProfileAttrs dir));
```
blaggacao, line 70: I think that could be problematic as it _looks like_ it creates no-op overlays for shallow nodes.

That might interfere with how those are exported.

src/importers.nix
```diff
@@ -1,73 +1,84 @@
...
-      hosts = recImport {
-        inherit dir;
-        _import = base: {
-          modules = import "${toString dir}/${base}.nix";
-        };
-      };
+      hosts = lib.mapAttrs
+        (n: p: {
+          # Only get top-level .nix files or default.nix from directories
+          modules = getProfilePath { } p;
```
blaggacao, line 79: Similar concern as for overlays. In general, we probably shouldn't create no-op (although internal) datastructures, here.

tests/lib.nix
```diff
@@ -49,21 +39,22 @@ lib.runTests {
...
       };
       profiles = [ (./profiles) ];
     };
     expected = {
-      bar = [ (toString ./profiles/foo) ];
+      bar = [ (toString ./profiles/foo) (toString ./profiles/f.nix) ];
       allProfiles = [
         (toString ./profiles/foo)
-        (toString ./profiles/t)
+        (toString ./profiles/f.nix)
```
blaggacao, line 57: I think a test with `default.nix` would be in order, too.

blaggacao (approved):
This looks very promising! A general thought on naming: now we have kind of a unified importer that equally is aplicable to:

- overlays
- modules
- profiles
- hosts

Maybe we can think of a generic name (that does not have "profile" in it)?

src/importers.nix
Pacman99: I guess filtering might a good idea here then. I don't want to pick up random folders without any `default.nix` in them and I don't want to discourage random folders - with the same reasoning as ignoring any file that doesn't have a `.nix` extension.

src/importers.nix
Pacman99: I think this is ok here, `hosts.modules` doesn't get exported and empty attrsets don't do anything.

If we were to write a `importers.modules` that could filter out any non-path values.

tests/lib.nix
Pacman99: `foo` has a default.nix

src/importers.nix
```diff
@@ -1,73 +1,84 @@
...
 
+  getProfilePath = fallback: item:
+    if lib.isString item then item else fallback;
 in
 {
-  inherit pathsIn recImport mkProfileAttrs;
+  inherit mkProfileAttrs;
+
+  profileMap = list: map
+    (p: getProfilePath (throw "passed profile ${builtins.toJSON p} isn't a path") p)
```
blaggacao, line 63: Log-prefix?

`[digga]: ` -- or maybe see other projects what's the convention, here.

blaggacao (commented):
Another thing.

Pacman99:
> This looks very promising! A general thought on naming: now we have kind of a unified importer that equally is aplicable to:

I do like the original idea of "everything is a profile". I realize that conflicts with the other naming of `profiles`. So yeah there should be some other concept, so we can say "everything is a ..."

blaggacao:
gummy bear.

collectProfiles. (But overlays aren't)

parseFolder.

collectFolder.

collectLeaves.

Pacman99:
collectLeaves! I want to use that

src/importers.nix
Pacman99: wait actually this won't interfere with exports, since overlays exporter looks for the names of values in the overlay. So if the output is a empty attrset, there won't be anything to export.

src/importers.nix
Pacman99: Isn't this still true? all top-leve `.nix` files and folders with a `default.nix` in them will get collected.

Pacman99:
update: rename: mkProfileAttrs -> collectLeaves

nrdxp:
Shouldn't it be rakeLeaves? 😂

Pacman99:
Also this change isn't entirely backwards compatible since collection stops at `default.nix`. I think it should be ok since you'll get a pretty easy to debug `attribute doesn't exist` error.

blaggacao:
@nrdxp you're a poet programmer :smile:

src/importers.nix
blaggacao: But also groups of files within a named folder, no? ("Grouping" use case with shallow nodes)

tests/lib.nix
blaggacao: Sry, still on cel phone. :-/

src/pkgs-lib/tests/lib.nix
```diff
@@ -1,93 +0,0 @@
-{ pkgs, lib, ... }:
```
blaggacao, line 1: Why is this file deleted?

blaggacao (commented):
There is still a wonder in my head:

Does this code allow or disallow multi-level grouping?

- `foo.bar.thing1`, but also:
- `foo.thing2`

via:

```
./foo/bar/thing1.nix
./foo/thing2.nix
```

If it disallows it, why?


---

And as I write, we probably should built in a mechanism that guards against name clashes when groupings need to be traversed and we can have two things named `thing` in different groups that we can't properly name.

This reminds me of a use of `flattenTree` and reduce nested groupings into a "/" separated identifier.

```nix
final: prev: {
  "foo/bar/thing" = { };
  "quz/qux/thing" = { };
}
```

Hm? What do you think?

_My opinion: People should probably not overcomplicate their lives and use such facilities extensively, but we still ought to be formally correct._

src/pkgs-lib/tests/lib.nix
Pacman99: should have been deleted earlier, libtests is moved to the `tests` folder, so this isn't even called. Thought I would just do it in this commit.

Pacman99:
Just tested, your example is allowed and works just as you described it.

I don't entirely get the motive for the `/` separation. Isn't it better to just use attrsets? `flattenTree` is only there for the caching speed of the one level deep `packages` output.

blaggacao:
> Isn't it better to just use attrsets?

It would. 

I think I made a thinking mistake.

Ignore me, I guess it would be too convoluted to explain.

src/importers.nix
blaggacao: I read this as:

Only rake nix files that are at fhe root of this directory or directories that are one (1) level deep.

Though, it is totally recursing.

src/importers.nix
```diff
@@ -1,73 +1,84 @@
...
-        _import = base: {
-          modules = import "${toString dir}/${base}.nix";
-        };
-      };
+      hosts = lib.mapAttrs
+        (n: p: {
+          # Only get top-level .nix files or default.nix from directories
+          modules = getProfilePath { } p;
+        })
+        (rakeLeaves dir);
```
blaggacao, line 81:
```suggestion
    hosts = map (getProfilePath {}) (builtins.attrValues (rakeLeaves dir));
```

blaggacao (commented):
`hosts` implemented analogous to `overlays`, if that makes sense

src/importers.nix
Pacman99: but this doesn't work, `hosts` argument is an attribute set.

blaggacao: So I got it wrong. :-/

blaggacao (approved):
We decided to stop short of full nested grouping for hosts and overlays. Those features are present with `rakeFlakes` and available in profiles.

For hosts, we need to resolve naming in a nested scenario.

For overlays, we need to resolve flattening.

Those are addressed in follow up PRs.

PR #32, opened 2021-05-27
imp: make importers truely recursive

blaggacao:
- hostnames are construed from a reverse DNS notation of folder names

BREAKING CHANGE (without practical impact):

fup needs to be udpated to parse the reverse DNS notation back into
a hostname DNS label + a DNS domain (to set networking.domain).

However, since until now, no recursive hosts where allowed, this
can be safely done soonishly.

### DIFF
--- a/src/importers.nix
+++ b/src/importers.nix
@@ -1,5 +1,63 @@
 { lib, devshell }:
 let
+  flattenTree =
+    /**
+      Synopsis: flattenTree _tree_
+
+      Flattens a _tree_ of the shape that is produced by rakeLeaves.
+
+      Output Format:
+      An attrset with names in the spirit of the Reverse DNS Notation form
+      that fully preserve information about grouping from nesting.
+
+      Example input:
+      ```
+      {
+        a = {
+          b = {
+            c = <path>;
+          };
+        };
+      }
+      ```
+
+      Example output:
+      ```
+      {
+        "a.b.c" = <path>;
+      }
+      ```
+    **/
+    tree:
+    let
+      op = sum: path: val:
+        let
+          pathStr = builtins.concatStringsSep "." path; # dot-based reverse DNS notation
+        in
+        if builtins.isPath val then
+        # builtins.trace "${toString val} is a path"
+          (sum // {
+            "${pathStr}" = val;
+          })
+        else if builtins.isAttrs val then
+        # builtins.trace "${builtins.toJSON val} is an attrset"
+        # recurse into that attribute set
+          (recurse sum path val)
+        else
+        # ignore that value
+        # builtins.trace "${toString path} is something else"
+          sum
+      ;
+    
+      recurse = sum: path: val:
+        builtins.foldl'
+          (sum: key: op sum (path ++ [ key ]) val.${key})
+          sum
+          (builtins.attrNames val)
+      ;
+    in
+    recurse { } [ ] tree;
+
   rakeLeaves =
     /**
       Synopsis: rakeLeaves _path_
@@ -84,11 +142,9 @@ let
     in
     lib.mapAttrs f imports;
 
-  getProfilePath = fallback: item:
-    if builtins.isPath item then item else fallback;
 in
 {
-  inherit rakeLeaves;
+  inherit rakeLeaves flattenTree;
 
   mkProfileAttrs = builtins.trace ''
     The function, mkProfileAttrs, has been deprecated, you can now create profile sets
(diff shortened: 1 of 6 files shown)


src/importers.nix
```diff
@@ -1,5 +1,64 @@
...
+        "a.b.c" = <path>;
+      }
+      ```
+    **/
+    tree:
+    let
+      op = sum: path: val:
+        let
+          isAttrs = builtins.isAttrs or (builtins.typeOf val) == "set";
+          pathStr = builtins.concatStringsSep "." path; # dot-based reverse DNS notation
```
Pacman99, line 36: can we instead just use upstream flattenTree and use `mapAttrs'` to adjust the naming:
```
mapAttrs' (n: value: {
  name = builtins.replaceStrings [ "/" ] [ "." ] n;
  inherit value;
}) (flattenTree dir)
```

src/importers.nix
```diff
@@ -18,38 +77,36 @@ let
...
-      # Relative paths cause issues, so convert to string immediately
-      dir = toString dirPath;
-
       collect = file: type: {
         name = lib.removeSuffix ".nix" file;
-        value = let path = "${dir}/${file}"; in
+        value = let path = dirPath + "/${file}"; in
           if (type == "regular")
-            || (type == "directory" && builtins.pathExists "${path}/default.nix")
+            || (type == "directory" && builtins.pathExists (path + "/default.nix"))
```
Pacman99, line 99: To be clear these would still end up as strings, but their just being converted here now right?

src/importers.nix
```diff
@@ -18,38 +77,36 @@ let
...
       };
 
       files = lib.filterAttrs
         (file: type:
-          # Only include `.nix` files or directories
+          # Only rake `.nix` files or directories
           (type == "regular" && lib.hasSuffix ".nix" file) || (type == "directory")
         )
-        (lib.safeReadDir dir);
+        (builtins.readDir dirPath);
```
Pacman99, line 109:
```suggestion
        (builtins.readDir (toString dirPath));
```
I really think this is a good idea for safety, I don't believe readDir works well with paths. I think thats the function that creates new nix store paths when you pass it a relative path.

src/importers.nix
```diff
@@ -111,19 +166,13 @@ in
   overlays = dir:
     {
       # Meant to output a module that sets the overlays option
-      # Only get top-level .nix files or default.nix from directories
-      overlays = map (getProfilePath (_: _: { })) (builtins.attrValues (rakeLeaves dir));
+      overlays = builtins.attrValues (flattenTree (rakeLeaves dir));
```
Pacman99, line 169: nice! flattenTree makes sense here

src/importers.nix
```diff
@@ -111,19 +166,13 @@ in
...
     {
-      # Meant to output a module that sets the hosts option
-      hosts = lib.mapAttrs
-        (n: p: {
-          # Only get top-level .nix files or default.nix from directories
-          modules = getProfilePath { } p;
-        })
-        (rakeLeaves dir);
+      # Meant to output a module that sets the hosts option (including constructed host names)
+      hosts = (flattenTree (rakeLeaves dir));
```
Pacman99, line 175: Ok yeah looks like this wouldn't change existing setups, but it would allow you to define groups of hosts.
This has to be adjusted to the `modules = [ ]` format thats required for the `hosts` argument.

tests/default.nix
```diff
@@ -18,7 +18,7 @@ in
       buildInputs = [
         pkgs.nix
         (
-          let tests = import ./lib.nix { inherit pkgs lib; }; in
+          let tests = import ./lib { inherit pkgs lib; }; in
```
Pacman99, line 21: Good idea, the folder should look much cleaner with this

Pacman99 (commented):
Cool idea, I see now where `flattenTree` comes into the picture.

src/importers.nix
blaggacao: upstream flattenTree also filters for derivations, out use is to stop on `path` types, so it's actually slightly different.

Pacman99: gotcha, it would be good to try and petition to change that. Since its better to use `filterPackages` instead. And `flattenTree` could instead by generalized for attrsets.

blaggacao: In the end upstream `flattenTree` is made to look for derivations though, which also are attrSets so that wouldn't work, either. I have no good answer at the moment, other than probably passing a seive to flatten tree. But that seive would be 80% of the function, so that doesn't seem ideal, neither.

src/importers.nix
blaggacao: it already does, the output is:

```nix
{
  "com.domain.hostname" = ./com/domain/hostname.nix # or .../hostname
}
```

blaggacao: Oh got ya! :wink: the `[]`

blaggacao: Resolved in https://github.com/divnix/digga/compare/4eedb92fb2f3aec3f55094814c01a6b99f423bfb..8613ff18eb8d8a6b60e8ef96503b40868c65bf0a

blaggacao:
bors try

Radvendii:
Oh I like this! I didn't even realize it was possible to have `"a.b"` as an attribute name, i thought it would auto-expand to `a.b`

Pacman99 (approved):
Very creative idea for importers!
bors r+

PR #36, opened 2021-05-31
fix: flk usage of lock update

blaggacao:
After discussion with @Pacman99 
the used command did not update as intended in one instance locally on my machine.

I din't investigate further, but `nix flake lock` seems the correct way to to this.


Pacman99 (approved):
bors r+

PR #38, opened 2021-05-31
importers: filter out empty attrsets

Pacman99:
Allows you to have arbitrary folders that only get picked up if .nix files are in there. This could be useful for a folder dedicated to contains patches. And generally improves safety of the now fully recursive importers.

I decided to put this in rakeLeaves, since it makes no sense to rake pointless folders and the change propagates to all importers.

### DIFF
--- a/src/importers.nix
+++ b/src/importers.nix
@@ -48,7 +48,7 @@ let
         # builtins.trace "${toString path} is something else"
           sum
       ;
-    
+
       recurse = sum: path: val:
         builtins.foldl'
           (sum: key: op sum (path ++ [ key ]) val.${key})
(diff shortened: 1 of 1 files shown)


Pacman99:
bors try

blaggacao:
Do I see correctly this would prevent:
```
./a/b.foo
```
from getting picked up?

Pacman99:
Yes those would have ended up as:
```
a = { };
```
or
```
a = { b = { }; }
```
Which I think actively hurts more than helps. Will add test cases now.

Pacman99:
bors try

blaggacao (approved):

blaggacao:
bors r+

Nice catch!

PR #39, opened 2021-05-31
generate registry and add fup-repl to env

Pacman99:
fixes #17
The command is just `repl`, I don't know if it needs to be `flk repl` since its a really generic command.

PR #40, opened 2021-05-31
flake: use fup for non-lib outputs

Pacman99:
Should help simplify #39 this means we can now access `fup-repl` in diggas devshell too.

Also we may as well offload some of the pkgs creation work to fup.


blaggacao (approved):
LGTM

PR #43, opened 2021-06-03
Added vm and vm-run commands to flk.sh

### LINKED ISSUE #41: add flk vm command
I can't paste in the commands now, but following the manual from a clean OVA I had a few issues. Later I'll post out issues from each step.

## Your Environment
https://channels.nixos.org/nixos-20.09/latest-nixos-x86_64-linux.ova

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/97544767-nixos-virtualbox-ova-setup-instructions-do-not-work?utm_campaign=plugin&utm_content=tracker%2F161955272&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F161955272&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Narice:
Closes #41.

That was far easier than what I thought! bash scripting is really straight forward :smile: 
What this does:
- `flk vm HOST` generates the vm for the host under the vm folder, the link to the nix store result is named after the HOST.
- `flk vm-run HOST` generates the vm as `flk vm HOST` and runs the vm creating the disk image under the vm folder as `HOST.qcow2`

Here is the tree after doing a `flk vm-run NixOS`:
```
vm
├── NixOS -> /nix/store/<hash>-nixos-vm
└── NixOS.qcow2
```

This allows for fast iteration on the NixOS configuration which is super useful, I already love it! :heart:

### DIFF
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -100,6 +100,7 @@ case "$1" in
         && "$DEVSHELL_ROOT/vm/tmp/$3/bin/run-$3-vm" \
       ) \
       && rm -rf "$DEVSHELL_ROOT/vm/tmp/$3"* \
+      && rmdir --ignore-fail-on-non-empty "$DEVSHELL_ROOT/vm/tmp"
     else
       nix build \
         "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.vm" \
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -102,6 +102,8 @@ case "$1" in
       "$DEVSHELL_ROOT/vm/$2" \
       "${@:3}" \
       && \
+      rm -f "$DEVSHELL_ROOT/vm/$2.qcow2" \
+      && \
       (export NIX_DISK_IMAGE="$DEVSHELL_ROOT/vm/$2.qcow2" \
       && \
       "$DEVSHELL_ROOT/vm/$2/bin/run-$2-vm" \
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -21,6 +21,8 @@ usage () {
   "get (core|community) [DEST]" "Copy the desired template to DEST" \
   "doi HOST" "Generate DigitalOcean image of HOST" \
   "iso HOST" "Generate an ISO image of HOST" \
+  "vm HOST" "Generate a vm for HOST" \
+  "vm-run HOST" "Generate and run a vm for HOST" \
   "install HOST [ARGS]" "Shortcut for nixos-install" \
   "home HOST USER [switch]" "Home-manager config of USER from HOST" \
   "HOST (switch|boot|test)" "Shortcut for nixos-rebuild" \
@@ -83,6 +85,27 @@ case "$1" in
     nix build \
       "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" \
       "${@:3}"
+    ;;
+
+  "vm")
+    nix build \
+      "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.vm" \
+      -o \
+      "$DEVSHELL_ROOT/vm/$2" \
+      "${@:3}"
+    ;;
+
+  "vm-run")
+    nix build \
+      "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.vm" \
+      -o \
+      "$DEVSHELL_ROOT/vm/$2" \
+      "${@:3}" \
+      && \
+      (export NIX_DISK_IMAGE="$DEVSHELL_ROOT/vm/$2.qcow2" \
+      && \
+      "$DEVSHELL_ROOT/vm/$2/bin/run-$2-vm" \
+      )
     ;;
 
   "install")
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -22,7 +22,7 @@ usage () {
   "doi HOST" "Generate DigitalOcean image of HOST" \
   "iso HOST" "Generate an ISO image of HOST" \
   "vm HOST" "Generate a vm for HOST" \
-  "vm-run HOST" "Generate and run a vm for HOST" \
+  "vm run HOST" "run a one-shot vm for HOST" \
   "install HOST [ARGS]" "Shortcut for nixos-install" \
   "home HOST USER [switch]" "Home-manager config of USER from HOST" \
   "HOST (switch|boot|test)" "Shortcut for nixos-rebuild" \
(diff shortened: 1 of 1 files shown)


Narice:
When booting multiple times, sometimes it doesn't work and I don't know why. There are failed jobs by systemd and I cannot login to the nixos user for example.
deleting the qcow2 fixes the problem, so I will delete the file if using vm-run for now.
I will precise it in the help documentation as well so that it is clear.
the vm-run is here to test a config and not to build a vm so it seems pretty fair to do that, what are your thought on this?

src/pkgs-lib/shell/flk.sh
```diff
@@ -77,6 +79,29 @@ case "$1" in
...
 
+  "vm")
+    nix build \
+      "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.vm" \
+      -o \
+      "$DEVSHELL_ROOT/vm/$2" \
+      "${@:3}"
+    ;;
+
+  "vm-run")
```
Pacman99, line 90: We could do something similar to `flk home` which checks if the third argument is `switch` and builds and switches. So that way its either `flk vm` or `flk vm run`.

Pacman99:
>  the vm-run is here to test a config and not to build a vm so it seems pretty fair to do that, what are your thought on this?

Yeah I would say if you do `vm run` it starts the vm and everything can get cleared after your done. Its the same thing we do for `flk home switch`, so the process would be consistent with that.

Pacman99:
> When booting multiple times, sometimes it doesn't work and I don't know why. There are failed jobs by systemd and I cannot login to the nixos user for example.

Does using the vmWithBootLoader build change anything?

Narice:
Ok I will try to do that, if it's a vm run then I guess I do it in a tmp folder and then rm the tmp folder once it's done.
We wouldn't need the vmWithBootLoader if we do that as the qcow2 gets cleared everytime, that was the problem so no bootloader is needed, startup will be faster I guess without it :wink:

Narice:
Do I also have to rebase this PR on develop @Pacman99?

Pacman99:
> Do I also have to rebase this PR on develop @Pacman99?

Ohh right yes please, I forgot to switch this one.

> Ok I will try to do that, if it's a vm run then I guess I do it in a tmp folder and then rm the tmp folder once it's done.

I don't think a tmp folder is necessary, you can just use the current directory, since the `vm` folder is going to be remove after the vm closes anyways. Or maybe theres a way to get `nix build` to just output a store path, but not symlink the result. Then that could just be stored in a variable in run. If thats possible perhaps `flk home switch` needs to be updated too.

Narice:
> Ohh right yes please, I forgot to switch this one.
Done!

> I don't think a tmp folder is necessary, you can just use the current directory, since the `vm` folder is going to be remove after the vm closes anyways. Or maybe theres a way to get `nix build` to just output a store path, but not symlink the result. Then that could just be stored in a variable in run.

The vm folder is not removed when the vm closes, if the person wants to simply build the vm, then it will be in the vm folder and stay there. If the person wants to save the vm he created, it would be bad to delete it after a `vm run`.
So that is why the tmp directory or file(s) (qcow2 + nix store link(?)) here is important.

What we can say in the documentation is this:
- vm is for generating a vm, run it under vm/HOST/bin/run-HOST-vm
- vm run is for running a one shot vm

that would be simpler for the user.
We would however have to state in the documentation how to set where the host image is in this case.

there is this for having the store path at the end and having no link:
```
nix build ... --no-link --json
```
however it gives this:
```
[{"drvPath":"/nix/store/wqkxfpwfvgkzhhb5yl3n42q85bsdmjbk-nixos-vm.drv","outputs":{"out":"/nix/store/3makb49117kajgfd71jiwkap8984myhj-nixos-vm"}}]
```
so that is not really good if we don't include as a dependency of the devos shell `jq` for parsing it :/

Narice:
I tried vmWithBootLoader for the vm command and no. it has the same problem than without bootloader.
It seems that in certain cases there is a chown problem:
image
and then it seems to snowball:
image

From the first screen we can see that is it a problem with the execution of a user script.
Would that be an upstream issue?

Steps to potentially reproduce it:
- run the vm
- when arriving at the login section, quickly close the vm (not a normal shutdown)

When using the vm like a machine using the shutdown, in that case from my testing, there is no problems.

Narice:
Just for reference, the only way I have found to get the machine back to a working state is to delete the `.qcow` disk image

Narice:
Normally the `flk vm` command should be really nice now.
I fusionned the `vm` and `vm-run` commands as suggested. I also use a tmp folder when making a `vm run`.
On exit of the one shot vm, I delete the vm files in the tmp folder and delete the tmp folder if it is empty.
That enables side-by-side opening of different hosts without conflict between the different commands executed.

On `flk vm HOST` I create a script file for running the vm with the right disk image file.

src/pkgs-lib/shell/flk.sh
```diff
@@ -21,6 +21,8 @@ usage () {
   "get (core|community) [DEST]" "Copy the desired template to DEST" \
   "doi HOST" "Generate DigitalOcean image of HOST" \
   "iso HOST" "Generate an ISO image of HOST" \
+  "vm HOST" "Generate a vm for HOST" \
+  "vm run HOST" "run a one-shot vm for HOST" \
```
Pacman99, line 25:
```suggestion
  "vm [run] HOST" "Generate a vm for HOST" \
```

src/pkgs-lib/shell/flk.sh
```diff
@@ -85,6 +87,31 @@ case "$1" in
       "${@:3}"
     ;;
 
+  "vm")
+    if [[ "$2" == "run" ]]; then
+      rm -rf "$DEVSHELL_ROOT/vm/tmp/$3"* \
+      && nix build \
```
Pacman99, line 93: I'm not very proficient in bash scripting, but is the `&&` necessary for every line here? I think the bash script would fail if any line fails.

Pacman99 (approved):
This looks great already. Its ok if there are some issues with the vm, that could also be an upstream issue. Incremental changes are good too. Go ahead and merge when you think its ready(with a bors r+)
bors delegate+

Pacman99:
What we can say in the documentation is this:

>    vm is for generating a vm, run it under vm/HOST/bin/run-HOST-vm
    vm run is for running a one shot vm

Yeah the reasoning makes sense to me here, once you start the vm, if its good you can access it in the `vm` folder.

src/pkgs-lib/shell/flk.sh
Narice: yes, if at one step something goes wrong I want it to stop, it would be cool if there was an error message for each step but I don't really know how to do that, that seems like an incremental step that could be done :wink:

Pacman99: I thought that was the default for bash, if one line fails, the bash script stops.

src/pkgs-lib/shell/flk.sh
Narice: here I wanted to convey the fact that the vm run HOST command is very different from the vm command as it does not save the vm, the vm is destroyed at the end.
The documentation could be written another way on one line without losing on the documentation detail but I don't really know how as of now, I will leave it as an incremental change to be done afterwards.

Pacman99: Got it, that makes sense

Narice:
I'm going to merge the pull request but wanted to know if issues should be created to keep track of the possible incremental changes brought up during this conversation?
- Making error messages for each step of the vm / vm run process
- Making the documentation for vm a one liner (or a help menu could be created for the vm subcommand)
- For the vm script and one-shot vm, getting the nix store path created from the nix build instead of making a link to it

bors r+

Pacman99:
Great! thanks for the PR, an `flk vm` is very exciting.

I would say one tracking issue would be good for improving `flk vm` overall.

Narice:
> I would say one tracking issue would be good for improving `flk vm` overall.
Alright! I'll do that then ^^

blaggacao:
I think here, too we need a modification to `devos`'s `.gitignore` (in cas you havn't had that already on the radar).

:+1:

PR #45, opened 2021-06-03
Changed output from nix build for iso and doi command

### LINKED ISSUE #44: `flk iso HOST` should put the link to the iso in an iso directory at the root of the devshell
`flk iso HOST` should put the link to the iso in an iso directory at the root of the devshell.
That makes sure that the iso file is at a specific place and we can build on that.
I will try to make a pull request for it.

Narice:
Closes #44.

This will make the command easier and more predictable.
We can also as a result add the iso folder in the devos gitignore.

### DIFF
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -82,6 +82,8 @@ case "$1" in
   "iso")
     nix build \
       "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" \
+      -o \
+      "$DEVSHELL_ROOT/iso/$2" \
       "${@:3}"
     ;;
 
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -76,6 +76,8 @@ case "$1" in
   "doi")
     nix build \
       "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.digitalOcean" \
+      -o \
+      "$DEVSHELL_ROOT/doi/$2" \
       "${@:3}"
     ;;
 
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -85,7 +85,7 @@ case "$1" in
     nix build \
       "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" \
       -o \
-      "$DEVSHELL_ROOT/iso/$2" \
+      "$DEVSHELL_ROOT/iso/$2.iso" \
       "${@:3}"
     ;;
 


Pacman99:
For future reference we do all development on the `develop` branch, I just switched the base branch for you, but you'll need to rebase on develop.

I definitely agree, this makes the custom builds much more clear if you get a folder with what you expect. Can you also include a similar change for the recently added `doi` image.

src/pkgs-lib/shell/flk.sh
```diff
@@ -74,6 +74,8 @@ case "$1" in
   "iso")
     nix build \
       "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" \
+      -o \
+      "$DEVSHELL_ROOT/iso/$2" \
```
Pacman99, line 78:
```suggestion
      "$DEVSHELL_ROOT/iso/$2.iso" \
```
Would that make sense? or is it redundant since the file is already in a iso folder?

Narice: Actually the file is just a symlink and it is right now just `result` but I agree, it is probably clearer.
What do you think it should be for a symlink, is there a standard for that?

Narice:
> For future reference we do all development on the `develop` branch, I just switched the base branch for you, but you'll need to rebase on develop.

I thought about that but didn't know if it was master or develop, that is good to know :+1:

Narice:
I rebased the branch on develop. I did not do for now the suggested change as I would like your view on what I added first.
Is that the last change you would like to see on this PR?

I'm planning to do the doi one in another PR for clarity but should I put it here and change the title?

src/pkgs-lib/shell/flk.sh
Pacman99: I just tested `nix build -o`(didn't know that was a thing), I would think `flk iso NixOS` would output as `iso/NixOS`. Is that right? Because if thats true then I think the `.iso` is unnecessary. If it outputs as `iso/result` then thats a little weird.

Pacman99:
>  Is that the last change you would like to see on this PR?

Yeah I just want to understand the output format.

> I'm planning to do the doi one in another PR for clarity but should I put it here and change the title?

I think I would prefer that, to just get both done in one PR. But If you prefer a separate PR thats fine.

src/pkgs-lib/shell/flk.sh
Narice: yes, it ouputs as `iso/NixOS` instead of `./result` which is way better.
If you want to, I can change it to `iso/NixOS.iso`, it's a matter of standard more than anything, should a link to a file respect the extension of the pointed file?

Pacman99: Ok perfect.
> If you want to, I can change it to iso/NixOS.iso, it's a matter of standard more than anything, should a link to a file respect the extension of the pointed file?

Good point, I think theres also a argument for clarity that having the `.iso` extension is more obvious. And some gui iso flashers might only show .iso files in the file selector.

So yeah I vote for a `.iso` extension for the file, even though its a symlink.

Narice: > And some gui iso flashers might only show .iso files in the file selector.
Oh! that's a really good point, I'll change it, yes .iso should be clearer as well ^^

Pacman99 (approved):
Thank you! I like the "more obvious" builds.
bors r+

Narice:
Do you happen to know the extension of the file built by the doi command @Pacman99?

Pacman99:
> Do you happen to know the extension of the file built by the doi command @Pacman99?

Ohh good point, it seems to be a `.qcow2` image, looking at `nixos/modules/virtualisation/digital-ocean-image.nix`. I'll push a commit to update the extension.

Narice:
> I'll push a commit to update the extension.

Ok, thank you! :grin:

blaggacao:
I guess we need to make sure to add those to the .gitignore in `devos`. Just in case you havn't already... :wink: 

Big thanks for this!!! :clap:

Narice:
Ah yes that is right! I will try to make a PR for it but if you beat me to it that's fine too!

Pacman99:
> Ah yes that is right! I will try to make a PR for it but if you beat me to it that's fine too!

I pushed a commit to devos develop already for it. I don't think theres any need for it in digga.

Narice:
Cool! thanks :+1:

PR #50, opened 2021-06-09
Add default customBuilds from nixos-generators

### LINKED ISSUE #49: Integrate with nixos-generators
We already have `flk iso|vm|doi`, but users want more. And we want to rely on the best available upstream aggregate for that task: [`nixos-generators`](https://github.com/nix-community/nixos-generators/pull/100) which shall become part of the [familiy](https://github.com/divnix/devos#family--like-family).

/cc @lassulus

Pacman99:
Idea is that we create an entry in `system.build` for every format in nixos-generators. Also introduce `config.lib.digga.mkBuild` so devos's customBuilds doesn't have to re-implement the logic.

closes #49 cc @Lassulus

### DIFF
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -77,8 +77,6 @@ case "$1" in
   "build")
     nix build \
       "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.$3" \
-      -o \
-      "$DEVSHELL_ROOT/builds/$2/$3" \
       "${@:4}"
     ;;
 
--- a/src/pkgs-lib/shell/flk.sh
+++ b/src/pkgs-lib/shell/flk.sh
@@ -19,8 +19,7 @@ usage () {
   "up" "Generate $DEVSHELL_ROOT/hosts/up-$HOSTNAME.nix" \
   "update [INPUT]" "Update and commit the lock file, or specific input" \
   "get (core|community) [DEST]" "Copy the desired template to DEST" \
-  "doi HOST" "Generate DigitalOcean image of HOST" \
-  "iso HOST" "Generate an ISO image of HOST" \
+  "build HOST BUILD" "Build a variant of your configuration from system.build"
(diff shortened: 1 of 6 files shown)


src/pkgs-lib/shell/flk.sh
```diff
@@ -75,20 +74,12 @@ case "$1" in
...
-      "${@:3}"
-    ;;
-
-  "iso")
-    nix build \
-      "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" \
-      -o \
-      "$DEVSHELL_ROOT/iso/$2.iso" \
-      "${@:3}"
+      "$DEVSHELL_ROOT/$3/$2.$3" \
```
Pacman99, line 81: Last remaining issue, I don't know how to infer the format extension. Right now as a hack I'm just using the name of the format, which is fine for `iso` but, for example, digital ocean would end up as `NixOS.do` instead of `NixOS.qcow2`.

Lassulus: some formats have an attribute formatAttr. which is a glob to find the desired output in the generated directory. maybe we add it everywhere and change it to always include the extension?

blaggacao: Example: https://www.github.com/nix-community/nixos-generators/tree/master/formats%2Fhyperv.nix

(`filname = "*.vdhx";`)

Arguably, I'd prefer a more explicit:

`extension = "vdhx";`

blaggacao:
```console
$ nix build --help
...
       • --out-link / -o path
         Use path as prefix for the symlinks to the build results. It defaults to result.
...
```

I read that this is only concerned with the "prefix". Wouldn't that give us a chance to have files written under that prefix (with whatever extension they want)?

blaggacao (commented):
LGTM.

People that need to name build results can pass `o/my/name`.

The very Nixpkgs upstream support is not in a shape that allows us to infer naming appropriately.

blaggacao (approved):

PR #56, opened 2021-06-14
add module importer (raker)

blaggacao:


tests/fullFlake/default.nix
```diff
@@ -45,10 +45,10 @@ let
       hostDefaults = {
         system = "x86_64-linux";
         channelName = "nixos";
-        modules = ./modules/module-list.nix;
+        imports = [ (lib.importers.modules ./modules) ];
```
Pacman99, line 48: Just a note that we're deviating a bit here from the way nixos works which maintains a module-list.nix. I think auto-importing is a friendlier default, so lets stick with this.

Pacman99 (approved):
Nice and intuitive addition :+1: 
bors r+

tests/fullFlake/default.nix
```diff
@@ -45,10 +45,10 @@ let
...
         externalModules = [
           { _module.args.ourlib = self.lib; }
-          ./modules/customBuilds.nix
+          ./modules/customBuilds.nix # avoid exporting
```
Pacman99, line 51: Ohh wait just realized, how do we avoid exporting this? won't customBuilds now be exported because of the auto-importing at the top?

Pacman99: although actually with #50, customBuilds is mostly safe to export because of no longer using `lib.builderArgs` we would just have to add the digga lib module to the export for.

PR #57, opened 2021-06-15
da/fix maybe import

blaggacao:
- fix: maybImport to also classic imports that peak for defualt.nix
- ref: don't allow strings for maybeImports, require proper paths objects


Pacman99:
The reason `maybeImportDevshellModule` is applied later is so that proper paths are passed to the modules from list importer, which needs them to infer module names. Thats why I extracted it as its own function and used it after argument evaluation.

blaggacao:
Oh I guess, I see.

> from list importer

which one is this?

blaggacao:
[Solution]https://github.com/numtide/devshell/blob/1f4fb67b662b65fa7cfe696fc003fcc1e8f7cc36/nix/importTOML.nix#L25) :smile:

Pacman99:
OH also this wont work because maybeImport never gets run on items in the `exportModuleType` again as thats supposed to be a list of paths.
Whats the reason for trying to refactor that right now? The current issue at hand can just be solved with updating `maybeImport` in evalArgs.

blaggacao:
Something appears to looks a bit like a code smell here, I'm not yet at the bottom of it, still diving :diving_mask:

blaggacao:
https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/72

src/mkFlake/evalArgs.nix
```diff
@@ -7,9 +7,16 @@ let
...
+      maybeImportDevshellModule = path:
+        if lib.hasSuffix ".toml" path then
+          devshell.lib.importTOML item
+        else import item;
+
+
       maybeImport = obj:
-        if (builtins.isPath obj || builtins.isString obj) && lib.hasSuffix ".nix" obj then
-          import obj
+        if builtins.isPath obj then
```
blaggacao, line 18: @Pacman99 wold the second commit actually be harmless?

Pacman99: Yes very harmful, it would disallow flake paths, which are referenced with strings, ex:
```
overlays = [
  inputs.agenix.overlay
];
```

Pacman99: Either way strings are a valid way to reference paths in nix, I'm very much against disallowing it.

Pacman99:
make sure to update fup here if you haven't already done so

blaggacao:
Thanks, did so force pushed in the first commit.

src/mkFlake/evalArgs.nix
blaggacao: OK, removed that change. I wasn't sure at all. :smile:

Pacman99 (approved):
Great cleanups, thank you!
bors r+

PR #59, opened 2021-06-19
move customBuilds out of the modules folder

blaggacao:


Pacman99:
I don't like adding more top-level files and customBuilds is a module. I'm guessing this change is to prevent exporting customBuilds, and I actually think we should just export customBuilds. The only thing that currently makes us unable to export it is the fact that it uses `config.lib.digga` so we need to find a way to include a module that adds `config.lib.digga` for customBuilds. Or alternatively perhaps `mkFlake` could just include `config.lib.digga` in every module export.

blaggacao:
Agreed!

PR #60, opened 2021-06-20
fixup! fix: maybe import well defined nix modules (with _file)

blaggacao:


blaggacao:
Going to merge in order to unblock downstream. Any oversight that should reveal itself, we can fix later.

PR #63, opened 2021-06-21
da/update nixos generators

blaggacao:
- update: nixos-generator with nixlib dep


Pacman99:
I think this needs a rebase and can we drop all the `follows` for nixos-generators. Using `follows` in a library flake breaks things. Until the nix PR fixing `follows` is merged, we shouldn't use it here.

blaggacao:
bors try

blaggacao:
bors try-

blaggacao:
bors try

flake.nix
```diff
@@ -10,15 +10,10 @@
...
-      # We only use the nixosModules output which only needs nixpkgs lib
-      # TODO: don't pull another 'nixpkgs' when only nixpkgs lib is needed
+      # Only nixosModules is used, so nixpkgs and utils isn't a cost
       nixos-generators = {
         url = "github:nix-community/nixos-generators";
-        inputs = {
-          nixpkgs.follows = "nixpkgs";
-          utils.follows = "utils";
-        };
+        inputs.nixlib.follows = "nixlib";
```
Pacman99, line 16: We need to remove this line too, otherwise `nixlib` will have to be an input for digga users.

Pacman99 (changes requested):

flake.nix
blaggacao: Why do you suggest that? This is an actual and accurate dependency.

Pacman99: Because of the nix issue... every downstream consumer of digga now has to add nixlib as an input. And of they don't their going to get a confusing error of how digga follows a nonexistent input nixlib.

Pacman99: We really just cant use follows in digga or any library flake until https://github.com/NixOS/nix/pull/4641 is merged.

blaggacao: So basically you're saing that https://github.com/divnix/digga/pull/71 is not an appropriate solution?

Pacman99: Yeah essentially. I don't think theres much advantage anyways with using follows here. nixlib is so small that it won't matter. I get that its more "correct", but I think we should just wait for the fix upstream before using it here.

blaggacao: Maybe we should help puh the upstream issue, since it's arguably quite an annoying one. :shrug:

Pacman99: Well I currently consider the use of `follows` in digga to be a major bug, it prevents people from using digga develop because of the error. So is it ok to just remove it for now?

blaggacao:
This is superseded by #79

PR #67, opened 2021-06-27
flk tool

blaggacao:
- mv `flk` to its own subdir
- flk: create flk shell and split subcommands into files
- fix: usage of reverse-dns hostnames in flk scripts
- fix: flake functor with rebinding self

### TODOs

- [ ] pass a generalised reference to `selfFlake` that works in both, devshell and as a nixos `systemPkgs`


blaggacao:
superseded by https://github.com/divnix/flk

PR #69, opened 2021-06-30
devshell: remove nix and setup NIX_CONFIG

nrdxp:
Users may find it surprising to have a different Nix in the shell. This ensures they still have the proper features and caches setup when they enter the DevOS devshell. In addition, this also fixes shell completion.


blaggacao:
I think we can't avoid [incorporating patches into `nix`](https://github.com/divnix/digga/pull/71). Even though _if_ we do that, we are left with the "boostrapping" problem that you describe.

nrdxp:
Instead of dropping nix in the shell path, perhaps we could rename it to avoid conflicting with system installed nix, which would be preferable after DevOS has already been installed. Maybe `nix-bootstrap` or `nix-bs` for more fun :laughing:. Alternatively, we could simply reexport the required Nix as a flake output, so once one enters the shell, get's their NIX_CONFIG var setup, they can call n run .#nix?

blaggacao:
> Instead of dropping nix in the shell path, perhaps we could rename it to avoid conflicting with system installed nix,

I like the idea! We could place it under `flk` as `flk nix` ?

nrdxp:
> Is that superior to nixConfig in flake.nix ?
Well according to the still remaining [FIXME](https://github.com/NixOS/nix/commit/343239fc8a1993f707a990c2cd54a41f1fa3de99#r44356843) upstream it is. Plus, you cannot add trusted keys with `nixConfig` still, even if it were fully working, whereas NIX_CONFIG seems to have no problem with them in my testing.

blaggacao:
This was essentially implemented via https://github.com/divnix/digga/pull/89

PR #70, opened 2021-06-30
mv flkShell to devos

blaggacao:
ref: split digga shell into divnix/flk & divnix/devos/shell (devshell)

companion of https://github.com/divnix/devos/pull/334


blaggacao:
bors try

blaggacao:
bors try

blaggacao:
bors try

blaggacao:
(bumped downstream test reference)

bors try

src/modules.nix
```diff
@@ -59,7 +59,6 @@
       ];
 
       # package and option is from fup
-      environment.systemPackages = [ pkgs.fup-repl ];
```
blaggacao, line 62: This needs to be made reproducibly available to `divnix/flk`, instead. (avoid "layer violation")

Pacman99: I would like this line to stay. `repl` is a tool that fup provides that digga installs by default. It is also reproducibly available in flk - or it was when I set it up. But thats separate from the fact that the `repl` command is a useful tool for your system.

blaggacao: If that's ok, I'll regard (new) https://github.com/divnix/flk/issues/5 as blocking for this PR, then. :+1:

blaggacao: Resolved via https://github.com/divnix/digga/compare/2a8e8a25057d82ed0539809e840c096f47e552e9..52c62dccaefba2ff501fcb05f1947e54cafbcdcf

blaggacao:
bors try

blaggacao:
That was the wrong downstream pin. Let's do it again with the correct one.

bors try

blaggacao:
bors try

blaggacao:
@Pacman99 & @nrdxp IMO, this is good to go.

This check tested the head of https://github.com/divnix/devos/pull/334, which, after merging this and updating the flake.lock there can be undrafted and merged as well.

That `devos` PR tip uses the current tip of `flk` which has seen quite some improvements and fixes several issues and regressions with `flk` that where present here on `develop` previous to this PR.

This does not update any docs, since they are all consolidated in `devos`. I think after this independization of `flk` it would be time to put the docs in their respective places, as well.

blaggacao:
bors try

(after rename to `bud`

blaggacao:
bors try

(after downstream bump)

blaggacao:
bors try

blaggacao:
bors try

(with a missing inputs follows fix & a effective workaround for `follows` bug)

Pacman99 (approved):

PR #71, opened 2021-06-30
imp: add proper nix patching

blaggacao:
It is very likely that some digga functionality requires patches to nix

Create a place for patches that mimicks the way how they could be
applied once `inputs.<name>.patches` of the
UFR (Unofficial Flakes Roadmap) is implemented.

Also showcase how `nixConfig` can add extra feature flags (without
overriding the patched-in ones).


Pacman99:
Im against adding a patched nix as the default for digga. Nix takes a while to compile for me and im certain for others. I don't want to wait almost an hour just to switch configurstions. I think the proper solution is to just not rely on the follows fix until its actually merged into upstream.

And since nixlib and deploys overlay the extra dependency isnt that big of an issue.

blaggacao:
> Nix takes a while to compile for me and im certain for others

Oh yeah, this is really, really annoying. I was hoping that maybe we could take a pinned `nix` inputs and then ask @nrdxp to provide a proper cached binary...

I guess we will be in need of some flexibility w.r.t. `nix` in the weeks to come. I don't possess proof, it's just a sentiment I have judging by the amount of people actually trying to fix things and the current upstream review capacity.

GTrunSec:
>I guess we will be in need of some flexibility 

Except for running lock --update-input, there is no reason to add it as default.

So, how about the patched-nix in commands of devshell.
```
 [[commands]]
   2 name = "patchedNix"
   1 command = "nix run github:<repo>#patchedNixUnstable -- $@"
   4 help = "Run Patched Nix"         
```

Pacman99:
> Except for running lock --update-input, there is no reason to add it as default.

This is a good point, but I think you would still run into an issue with nix complaining about an input following another non-existent one during other calls of `nix`, because nix reads the lock file on every call.

blaggacao:
>  input following another non-existent one during other calls of nix, because nix reads the lock file on every call.

As far as my experiemtns got me during the last couple of days, a well-crafted flake.lock file can be consumed by a non-compliant `nix` as well. That is: the file is correctly read. The only issue is while updating.

blaggacao:
bors try

blaggacao:
bors try

blaggacao:
Note this only exposes what's already used for the devshell. It's up to the consumer to use it.

PR #72, opened 2021-06-30
imp: conservative digga lib output

blaggacao:
in order to avoid people accidentially depending on some internal API
which we don't explicitly want to support well into version 1.0, hide
those from the output.

Users who still want to use those need to copy them over, instead.

This reduces our API exposure and hopefully will help save maintainer
resources once `digga` completely takes over the world.


blaggacao:
bors try

Pacman99 (approved):
Yes! We're removing attrs.nix/lists.nix/strings.nix, I've been wanting to do that

PR #73, opened 2021-07-03
use deploy's overlay to prevent nixpkgs dep

Pacman99:


flake.nix
```diff
@@ -4,17 +4,13 @@
   inputs =
     {
       nixpkgs.url = "github:nixos/nixpkgs";
-      deploy = {
-        url = "github:serokell/deploy-rs";
-        inputs = {
-          nixpkgs.follows = "nixpkgs";
```
Pacman99, line 10: We should avoid using `follows` since it breaks downstream due to the nix issue. And theres really no use in doing it anymore, because nixpkgs isn't going to be pulled and utils is so small.

Pacman99:
bors try

blaggacao:
Nice! Thanks for not touching the shell w.r.t. https://github.com/divnix/digga/pull/70 :smile:

flake.nix
blaggacao: I'm not sure, but it could be that `follows` might become the preferred way of homologating dependencies in the long run. Not `overlays` which is local to `nixpkgs`. Sure, as long as everyone takes dependency on `nixpkgs` for granted, this issue does not yet surface.

Anyway, that's a long shot into the future and not relevant for this PR.

blaggacao (approved):

Issue #76, opened 2021-07-06
`mkFlake` assumes valid values for `nixos.hostDefaults.channelName`


blaggacao:
[`mkFlake` requires `nixos.hostDefaults.channelName` to be set](https://github.com/divnix/digga/blob/adee82d42dfcf6c34ccc8a2303116b17750bf522/src/mkFlake/default.nix#L99) to a meaningful value. But that is not enforced (`default = null;`)

Pacman99:
closed by #77

PR #77, opened 2021-07-06
fix: make nixos.hostDefaults.channelName required

blaggacao:
In several places `mkFlake` assumes an valid value is set, here.

fixes: #76

- [x] A nicer error message:
```
nix develop
warning: Git tree '/home/blaggacao/ghq/github.com/blaggacao/devos' is dirty
error: The option `nixos.hostDefaults.channelName' is used but not defined.
```

### DIFF
--- a/src/mkFlake/default.nix
+++ b/src/mkFlake/default.nix
@@ -149,7 +149,7 @@ lib.systemFlake (lib.mergeAny
                     };
                   };
 
-                  profilesTests = 
+                  profilesTests =
                     # only for hosts that also are the same system as the current check attribute
                     if (hostConfigsOnThisSystem != [ ])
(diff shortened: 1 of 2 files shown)


blaggacao:
As an alternative, we could set the `channelName` per API container, where it's required. Maybe that's better?

Pacman99:
> As an alternative, we could set the channelName per API container, where it's required. Maybe that's better?

Maybe, but the fact that its necessary to differentiate the two makes me feel like a change in mkFlake/stripHosts is warranted.

Pacman99 (approved):
Ok yeah I think this is the best solution. Let do it as a function like this so that its clear to code readers why the distinction is made in the two api containers.

blaggacao:
BTW, I concluded that `emptyValue` would not help us in any way here. It would simply fail. I tested.

Issue #78, opened 2021-07-06
Streamline inputs


blaggacao:
- pin nixpkgs to release-21.05 (chances are downstream users are using this as their default channel, so should we ever provide a cache here, they will hit it)
- express dependency on `nixlib` vs `nixpkgs` via an alias of `nixlib` to `nipkgs` (which effectively is a superset of `nixlib`). Thereby we save on one network fetch, while at the same time not foroing conciseness and precision while delcaring inputs)

blaggacao:
Closed via https://github.com/divnix/digga/pull/79

PR #79, opened 2021-07-06
imp: reduce flake inputs footprint and increase conciseness

blaggacao:
also produce more cach hits using nixpkgs/release 21.05

fix: #78


Pacman99:
Why are we adding more use of `follows`? This means every user of digga has to add nixlib, utils, and nixpkgs to their input. Which make no sense since their flakes don't directly depend on those things, only digga does.

blaggacao:
The goal of this PR is:
- to improve conciseness of the inputs section and also the its `flake.lock` file
- to pin `nixpkgs` to `release-21.05` to increase chances of cache hits throughout the different inputs used.

flake.nix
```diff
@@ -3,27 +3,20 @@
...
-      nixpkgs.url = "github:nixos/nixpkgs";
-      deploy = {
-        url = "github:serokell/deploy-rs";
-        inputs = {
-          nixpkgs.follows = "nixpkgs";
-          utils.follows = "utils";
-        };
-      };
+      nixpkgs.url = "github:nixos/nixpkgs/release-21.05";
+      nixlib.follows = "nixpkgs"; # "github:nix-community/nixpkgs.lib";
```
Pacman99, line 7: Whats the reason for this line?

blaggacao: The concisely expressed usage intent is `nixlib`. However, meanwhile `nixpkgs` is a strict superset of `nixlib` and meanwhile we also happen to depend on `nixpkgs`, we can alias `nixpkgs` to `nixlib` instead of using proper `nixlib`.

For conciseness sake, I could have left out the commented link, but having it there helps people understand what is the shape of (the relatively new) `nixlib` that is intended and they can conclude the alias is due to `nixpkgs` being a superset of it.

blaggacao:
bors try

flake.nix
blaggacao: @Pacman99 Did I answer your question? And is it clear how this serves the conciseness goal of this PR?

blaggacao:
bors try

PR #82, opened 2021-07-08
imp: use `deploy-rs` to deploy hm activation scripts

blaggacao:
deprecate mkHomeConfigurations


Pacman99:
I would prefer deploy-rs as an alternate method. I'm against deprecating mkHomeConfigurations because its nice to be able to use `home-manager switch`. I don't see the issue with supporting two methods of installing configurations, we do it for nixos.

src/generators.nix
```diff
@@ -1,5 +1,45 @@
...
+              system = {
+                user = "root";
+                path = activate.nixos config;
+              };
+            }
+            //
+            (
+              builtins.mapAttrs
+                hm-profiles
+                config.home-manager.users
```
Pacman99, line 33: We should use the same pattern as `homeConfigurations` here, `deploy.nodes.user@host`

blaggacao:
> I don't see the issue with supporting two methods of installing configurations, we do it for nixos.

This is the main reason for roughly 50% of `bud` scripts. I definitely want to do away with them.

In more general terms, supporting two alternative methods for no good reason looks like making this unmaintainable real quick.

src/generators.nix
blaggacao: Remember that the given structure is:

`deploy.nodes.<host-reverse-dns>.profiles.<user>` probably you saw `deploy.nodes.<user>`?

Pacman99:
I feel like your pushing for a high ideal without thinking of practical usability. `homeConfigurations` is an official output for home-manager configurations and I have tested that to work for activating hm configs. I treat deploy-rs as a cool experiment, but not super well tested. And I would like my configurations to follow the widely used standard for exporting home-manager configs. 

My understanding for `bud`/`flk` was to just be a thin wrapper until the official tools were better. This PR seems to be going in the opposite direction of that.

src/generators.nix
Pacman99: Got it, that makes sense, yeah I misunderstood the structure

Pacman99, line 33:
```suggestion
                config.system.build.homes or config.home-manager.users
```
We should use the custom build here too, since that adds in the registry and allows for hm-only configuration

blaggacao:
> `homeConfigurations` is an official output for home-manager configurations

Oh, that changes things. So we can consume `self.homeConfigurations`, instead, and we can probably do away with the `customBuild` stuff in `devos` w.r.t. hm, is that correct?

Pacman99:
> Oh, that changes things. So we can consume self.homeConfigurations, instead, and we can probably do away with the customBuild stuff in devos w.r.t. hm, is that correct?

The customBuild stuff in devos is pretty necessary, a `home-manager` switch without it or even a `deploy` without it will fail, because you need to to set `home-manager.useUserPackages` to false. And theres the QOL improvement of being able to do `nix shell nixos#vim`.

blaggacao:
> The customBuild stuff in devos is pretty necessary, a home-manager switch without it or even a deploy without it will fail, because you need to to set home-manager.useUserPackages to false.

I'm not sure if I follow: does this mean `home-manager.useUserPackages` cannot be set in a regular module? I really apologize if I'm missing something obvious here, since I'm not too familiar yet with hm.

src/generators.nix
blaggacao: This is better resolved via https://github.com/divnix/digga/pull/84

blaggacao:
bors try

PR #83, opened 2021-07-08
Revert "Revert "use deploy's overlay to prevent nixpkgs dep""

Pacman99:
This reverts commit adee82d42dfcf6c34ccc8a2303116b17750bf522.

The reason for the original revert makes no sense, the re-evaluation of nixpkgs happens either way, either it can happen here or it happens in deploy(remember system-spaced lib outputs means an extra evaluation). I think doing `.appendOverlays` is cheaper than doing a full re-eval. But more importantly this change means digga isn't dependent on nixpkgs, only nixlib.

And we can either error out from a use of `follows` here or a use of `follows` in deploy-rs, I'd prefer it be the latter.


blaggacao:
I'd like to generate more insight on this. I havn't seen this problem on the original implementation:

To me, it appears that this adds an instantiation for every config hence penalizing quite a bit the use of `deploy-rs`.

On the other hand deploy-rs does not depend on any such marginal instantiation since the reason for it's dependence is on a couple of trivial-builders.

Pacman99:
> To me, it appears that this adds an instantiation for every config hence penalizing quite a bit the use of deploy-rs.

You're right, I missed that. I updated the PR to only evaluate it once by just taking an arbitrary nixpkgs from one of the hosts. It doesn't matter what we use, we just need a pkgs set with deploy's overlay

src/generators.nix
```diff
@@ -19,18 +19,21 @@
       Generate the `nodes` attribute expected by deploy-rs
       where _nixosConfigurations_ are `nodes`.
       **/
-
+    let
+      pkgs = (builtins.head (builtins.attrValues hosts)).config.pkgs.appendOverlays [ deploy.overlay ];
```
blaggacao, line 23: That looks a lot better.

We should clarify the arbitrary / "layer violating" character of using this idiom. I just want to make sure that it's clearly marked as what it is: "not clean".

As soon as cleanliness does not suffer any disadvantage (that is: the the `nix` follows fix is incorporated) we should then proceed to revert this.

Pacman99: I don't really agree with this. I would prefer using the overlay over using `follows`. With this method, we DO NOT depend on nixpkgs, only nixlib. If we use follows then digga does depend on nixpkgs. I like fup's idea of making use of the downstream consumer's nixpkgs, I would like to do that here too.

blaggacao: I think my point is:

`nixpkgs` is already not very concise. A hypothetical `trivial-builder` flake would be a lot more concise. But if this is true, then an arbitrary `pkgs` goes into the opposite direction and disguises even more the _actual_ dependency here.

Pacman99: I'm thinking of more practicality. If we use `.follows = "nixpkgs" instead of this overlay then digga depends on nixpkgs. By using the overlay we don't depend on nixpkgs. And because of the way system-spaced lib outputs works we can't have it follow anything other than nixpkgs. 

The idea here is we just need a nixpkgs, any instance will do.

blaggacao: > The idea here is we just need a nixpkgs, any instance will do.

Yep, that is completely true, until somebody downstream decides to depend on `import nixpkgs { config = {}; overlays = [];}` independently, which will get memoized, and hence is guaranteed to be evaluated only once, provided the different `nixpkgs`' input hashes conincide via use of `follows`.

Something like `import nixpkgs { config = {}; overlays = [];}` might even happen in a third party dependency where neither we nor downstream have full control over.

But as long as the `{ config = {}; overlays = [];}` contract would be honoured by such 3rd party tools, there is no marginal penalty in their use of `nixpkgs`, since we would already set the precedence, here.

I'd like to find a way to make this concoction of things transparent to the future code reader, when they read these line of code.

Pacman99: > until somebody downstream decides to depend on import nixpkgs { config = {}; overlays = [];} independently

I get that memoization is a useful tool, but I don't see how its better than just not depending on nixpkgs at all. And because deploy uses a nixpkgs instance with its own overlay the memoization won't even happen if we use deploys lib directly.

Pacman99: So I added a comment to explain trivial builders dependency and how any nixpkgs instance can be used. Those seem to be points we both agree on.

blaggacao: That looks good! Thanks!

blaggacao (approved):
LGTM

blaggacao:
The failure is probably my fault, will review it timely.

src/generators.nix
blaggacao: For the record: if other inputs start to honor the here postulated `{ config = {}; overlays = [];}` contract for the sake of memoization, then we can't make use of this and still add a then unnecessary marginal +1 evaluation due to the use of `appendOverlays`.

blaggacao:
Now that new error falls beyond my responsibilities :smile_cat:

blaggacao:
Could you please rebase on https://github.com/divnix/digga/pull/82 ? It would help save me from rebasing in tight coordination with `devos` three stacked PRs, which is every time a task worth 1-2 hours until all tests and input references are working again. Given the current PR inventory, I'd very much appreciate if that would be possible.

Pacman99:
> Could you please rebase on #82

I can't really do that, 82 code includes some code that might change soon(depending on homeConfigurations discussion). I think if the PR is only deploy stuff then I should be able to, but if portable homeConfigurations is tied to it then its difficult. This change is generally much simpler.

bors try

blaggacao:
Ok, let's wait with this PR until the PR inventory goes down a bit. I'm not really keen on doing this extra round of intricate rebases. And since those are conflicting PRs, let's apply "first come first serve".

All conflicting PRs are green and ready for review.

PR #84, opened 2021-07-09
ref: cleanup hm config generation

blaggacao:
- uses c.config.home-manager.users directly (depends on https://github.com/divnix/devos/pull/338)
- fix: uses c.config.networking.hostName instead of reverseDNS names
- sets `useUserPackages` to `false` and removes mkForce false from https://github.com/divnix/devos/pull/338)


fufexan (approved):

flake.nix
```diff
@@ -11,6 +11,8 @@
           utils.follows = "utils";
         };
       };
+      home-manager.url = "github:nix-community/home-manager";
+      home-manager.inputs.nixpkgs.follows = "nixpkgs";
```
Pacman99, line 15: This can be `nixlib`, since we're overriding the `pkgs` passed to homeConfiguration we don't need to incur nixpkgs dep here. But I guess we have to wait for follows fix.

blaggacao:
bors try

blaggacao:
bors try

blaggacao:
bors try

PR #85, opened 2021-07-14
hm users arg

blaggacao:
- ref: aclaratory naming
- imp: add portable per aarch users


blaggacao:
bors try

src/modules.nix
```diff
@@ -73,7 +73,7 @@
       };
 
       _module.args = {
-        inherit self;
+        inherit self hmUsers;
```
blaggacao, line 76: I concluded that it's preferable for `nixosConfig` to wrap `hmConfig`, rather than the other way round where `hmConfig` wraps an eventually available `nixosConfig`:

Not only does _eventual availability_ require the mandatory use of conditionals. And if a user forgets them, things _will_ break. But also would `hmConfig` wrapping `nixosConfig` be an inversion of the "control flow".

Therefore it is preferable to pass `hmUsers` to `nixosConfig` so that they can be activated like so:


```nix
{ hmUsers, ... }: {
  home-manager.users = { inherit (hmUsers) MYUSER; };
}
```

blaggacao:
bors try

PR #89, opened 2021-07-16
nixConfig: setup the nix configuration

blaggacao:
Users may not have setup their nix correctly. This ensures they have the
proper features and caches setup when they use this repo.

We currently have not entire clarity wether all `nixConig` flake options
work. At some point in time `nix` support was limited.

Therefore, we implement an NIX_CONFIG fallback (at least) in the devshell

supersedes #69


blaggacao:
bors try

PR #94, opened 2021-07-25
mkFlake: prevent export of empty outputs

Pacman99:


blaggacao:
Thank you! You beat me to it :smile:

blaggacao:
bors try

src/mkFlake/outputs-builder.nix
```diff
@@ -42,12 +45,14 @@ let
       config.home.users;
 
 in
-{
-
+lib.optionalAttrs (homeConfigurationsPortable != { }) {
   inherit homeConfigurationsPortable;
-
+}
+// lib.optionalAttrs (allOverlays != [ ]) {
```
blaggacao, line 51: I see a failure scenario, where `expuhtPackages` filters out 'invalid' packages due to `filterTree`. For example it filters out system mismatches. A possible solution would be to actually invoke `exportPackages` and only expose the result if not empty.

blaggacao (changes requested):

src/mkFlake/outputs-builder.nix
Pacman99: I tried to do that solution first but I ended up with infinite recursion because we pass self.overlays to exportPackages.

Pacman99:
I'm sort of stuck with this because its very difficult to check whether overlays and packages are empty without evaluating some functions twice, and doing that can cause infinite recursion. I'm pretty sure the goal here is possible, just requires some thinking.

I think for now maybe we could clean up exports for everything but overlays and packages.

PR #97, opened 2021-08-15
ref: add an option for profilesTests

GTrunSec:
This PR requirement from https://matrix.to/#/!UUqahLbShAYkkrXmKs:matrix.org/$haGVjkfZqLMa8LoXovUOy4Qjr_Ixl028HgzeMyx88pU?via=matrix.org&via=nixos.dev&via=nixos.org.


kraf1nix (approved):
## Motivation

Currently `digga` generates `allProfilesTests` for each `host`.
This works for hosts with simple configurations, however for hosts with multiple disks, secrets requirements and/or anything more complex these tests can fail, or pass (even with many services failing), so: 
* can increase testing/check time (needlessly)
* can pass when not necessarily a useful test for the `host` with many services failing
* can fail / timeout massively increasing test-time, or causing failures in `nix flake check`

For more complex hosts you will want to write custom tests to ensure functionality anyway.

## Result

- disable `allProfilesTests` for all hosts via `profilesTests.disableProfilesTests = true`
- disable `allProfilesTests` on a host-by-host basis with `profilesTests.removeHosts = [ "host-1" ];`

I have tested locally and confirmed that both the host filter + full disable work correctly :+1:

blaggacao:
I'd be much more inclined to seize the opportunity and remove the automatic tests all-together...

Since nobody really knows what they are doing, they might give a false sense of security.

Rather, we could provide a minimal reference implementation in `devos`.

w.d.y.t.?

GTrunSec:
>I'd be much more inclined to seize the opportunity and remove the automatic tests all-together...
Do you mean remove them from `flake check`?

>Rather, we could provide a minimal reference implementation in devos.
agreed. 
Also, decrypting one of the age keys would be a good example in customTests.

kraf1nix:
I think removing the automated tests and extending the custom tests would be a way to go.

> Also, decrypting one of the age keys would be a good example in customTests.

Agreed, I already have one of these set up with a snakeoil key for tests only, although I do it more so I can test the host (while injecting different ssh keys for agenix) rather than an agenix test.


If I had other requests for testing it would be:
* it would be nice to have access `lib` when defining tests in `devos`, when writing more complex/repeated tests this would be very useful
* (_this may be a bit of a work_): have a more global tests option where I could pull in multiple hosts so I could test the interaction between my hosts using the python testing framework's `nodes` with `nodes.host`
  * I think you could already achieve this by writing custom tests on a base host like `NixOS` and then moving full host configurations into `suites`, overriding `hostname` in each `nodes.host` and recreating `hosts` like that. Although that would require restructuring my code somewhat.
  * even in this approach I do like having `test` being an option of a nixos `host`, as you will still want to write tests specific to single hosts (in isolation).
  * this may be out of scope, or not the right way to approach multi-host testing

blaggacao:
Real quick, and undigested:

- The test option receives also a function that is basically `pkgs.callPackage`ed. so access to a `lib.mylib` should be doable. We could review the UX.

- Yeah, there should be a way to submit environment tests in addition to host tests.

blaggacao:
less is more & so is transparency better than black magic: https://github.com/divnix/digga/pull/101

PR #98, opened 2021-08-17
Update to latest follows fix patch sha

teutat3s:
Tested and Devos `flake.lock` works on latest `nixos` channel with this new patch sha.


GTrunSec (approved):

Pacman99:
I think we have to update the nix source too, so the patch applies cleanly

teutat3s:
> I think we have to update the nix source too, so the patch applies cleanly

Where would I do that? In the `flake.lock` file?

Pacman99:
I think bumping nixpkgs should be enough, lets see
bors try

teutat3s:
It seems this was already fixed via 2c5953f284690a89bfd472418141e3afea2dcf5d, feel free to re-open otherwise.

PR #99, opened 2021-08-31
Cleanup follows patch

teutat3s:
This PR proposes one way to clean up the follows patch, now that the corresponding PR NixOS/nix#4641 got merged.

One open question is: leave the patch code, but commented as an example?


blaggacao:
hm... we still need to ship the latest `nix` version though since some things depend on it, unless it's already in `nixpkgs`?

nrdxp:
latest version in nixpkgs is currently `2.4pre20210802_47e96bb`. I'll see if someone already opened a PR to bump it, otherwise I'll open one in a bit.

teutat3s:
https://github.com/NixOS/nixpkgs/pull/136271

flake.nix
```diff
@@ -129,7 +126,7 @@
       };
 
       # a little extra service ...
-      overlays = import ./overlays { inherit inputs; };
+      overlays = import ./overlays;
```
Pacman99, line 129: I would say we drop the export since patchedNix doesn't mean anything right now. I do agree with keeping the overlay file like you did so its easy to re-add patches.

blaggacao:
If the upstream bumps `nixUnstable` we still need to make sure that users somehow do "most likely" use the new version of `nixUnstable` since the CI is tested against 21.05 and that would be the base line guarantee.

Any ideas?

Pacman99:
Is the goal to check if users of digga are using 21.05? We could do that in the core module by adding an assertion or warning. We could even do the warning by checking for the nix version in `pkgs`.

flake.nix
teutat3s: With "drop the export" you mean removing `overlays = import ./overlays;` from flake.nix?

Pacman99: yeah exactly

blaggacao:
> Is the goal to check if users of digga are using 21.05?

We need to ensure for a while still that under all circumstances users run  `nix` post-follows patch on their systems to start pushing back on the associated issues a bit.

Although devos users have somewhat get used to the situation, pressing to ensure people run a fixed version (help-desk: "just place `digga.overlays.nix-patched` in your overlays - if that doesn't solve the problem, let's dig further").

blaggacao:
https://github.com/divnix/digga/pull/100/commits/79670b66387b918aab8854cae08f04fd0cc4608c#diff-e3de3cb458149cc755955e5a0634317073224f97c7568c30217f2d668ed2fa4bR6-R9 closed this.

@teutat3s thanks for pushing this!

PR #100, opened 2021-09-05
bump nix

blaggacao:
- fix: use soft-forked fup until upstream can schedule fixes
- bump: nix, use version that includes follows patch


blaggacao:
bors try

PR #101, opened 2021-09-05
imp: remove default application of profileTests

blaggacao:
now that we have per-host custom tests, it does not make sense anymore
to have magic lumpsum tests.

users are encouraged to use the new digga.lib.allProfilesTest standard
test and add it to a host's test like so:

`nixos.<host>.tests = [ digga.lib.allPrefilesTest ];`


blaggacao:
bors try

blaggacao:
@GTrunSec @kraf1nix should we adopt `devos` to implement by default `allProfilesTests`?

GTrunSec:
Sure, we should. The Devos does not have complex profiles for testing. Feel free to do that.

>(this may be a bit of a work): have a more global tests option where I could pull in multiple hosts so I could test the interaction between my hosts using the python testing framework's nodes with nodes.host

Next, we can be going to multi-nodes. Same idea, much simple much easy. Just pulling a profile to another node via `lib.suites.xxxx`

```
nixos.<host>.tests = [  
nodes = {
machine = { ... }: { };   
machine = { ... }: { machine = { ... }: { };};       
};
testScript = { nodes, ... }: ''
'';
];
```

blaggacao:
that should be already _almost_ possible: https://github.com/divnix/digga/issues/104 that code snippet there already haules other host's configs into `_module.args`. What would be an appropriate diff to `src/test.nix`?

PR #103, opened 2021-09-05
external not exported rename

blaggacao:
as the use case for digga expands into downstream
lib expantion the "external" naming is no more
appropriate when downstream lib functions purposefully
bundle and intentionally _re-export_ upstream art.

### DIFF
--- a/doc/api-reference-devshell.md
+++ b/doc/api-reference-devshell.md
@@ -19,13 +19,12 @@ submodule or path convertible to it
 
 
 
-## devshell.externalModules
-modules to include that won't be exported
-meant importing modules from external flakes
+## devshell.modules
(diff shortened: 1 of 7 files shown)


blaggacao:
bors try

blaggacao:
bors try

Pacman99:
I agree with the reasoning for the name change. I don't love `nonExportedModules`, I would prefer making this just `modules` and switching the current `modules` to `exportedModules`. It seems like modules have to be written well to be safely exported and such a name change would convey that better.

blaggacao:
I really like that reasoning!

blaggacao:
Would you mind taking over? i know I won't be able to carve out the slot to amend within the next X weeks.

PR #105, opened 2021-09-14
mkFlake/options: add pathToOr for modules options

Pacman99:
I still prefer the `modules = ./modules/module-list.nix` pattern, this makes that possible without having to use `import`.

### DIFF
--- a/doc/api-reference-devshell.md
+++ b/doc/api-reference-devshell.md
@@ -25,7 +25,7 @@ meant importing modules from external flakes
 
 
 *_Type_*:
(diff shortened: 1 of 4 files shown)


Pacman99:
bors try

blaggacao (approved):
This is neat.

I hope it does not break anything unexpected.

Pacman99:
I hope so too, I figure if overlays has the same thing, we should be ok. And we used to do this. I checked all the flake outputs too to make sure everything is there.

PR #107, opened 2021-09-24
rename current modules -> exportedModules

Pacman99:
and externalModules -> modules

supersedes #103

### DIFF
--- a/deprecated.nix
+++ b/deprecated.nix
@@ -90,6 +90,14 @@ rec {
       )
   ;
 
+  importModules =
+    lib.warn ''
+      Deprecated Function: lib.importModules.
+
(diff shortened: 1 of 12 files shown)


Pacman99:
The deprecation strategy here is just to error out if `externalModules` is used at all. There is no way to do any sort of "soft" deprecation scheme, because the renames overlap so in the module system trying to switch options will cause infinite recursion.

blaggacao (approved):

blaggacao:
bord r+

Pacman99:
bors try

Pacman99:
merging based on earlier approval and successful bors run

Issue #16, opened 2021-05-19
nix-darwin support


wangkev:
**Is your feature request related to a problem? Please describe.**
Not really. Would like to have `nix-darwin` support as well.

**Describe the solution you'd like**
Understand that this is currently targeted towards NixOS, but would be great to have support for [`nix-darwin`](https://github.com/LnL7/nix-darwin#flakes-experimental) builds as well.

**Additional context**
Unfortunately, I am just getting started with `flakes` (and generally `nix`), but think the philosophy of having a repository to get up and running on `nix` could apply across NixOS and macOS.

Pacman99:
Transferring to digga, this has to be implemented here first.

Pacman99:
Also I laid some initial groundwork for this a couple weeks ago here: https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/52. If anyone is interested in implementing this make sure to check that out.

PR #111, opened 2021-10-10
Add Support for Darwin Hosts

montchr:
To all curious observers: **this draft pull request is still a work in progress and not yet fully tested**. It does not yet work with the DevOS template, which will need updates to rearrange some NixOS-specific assumptions.

~I have not yet been able to get `nix develop` to succeed in any of the example flakes – it fails with an error every time, despite my efforts to make progress.~ Edit: builds are now passing consistently with the examples (except for downstream). 

I would appreciate any help – ~any insight into the cause of the error messages in CI would move this PR along.~ Edit: if you’re feeling adventurous, this needs testing for managing real macOS system configurations. 

You can see the current state of this PR's builds on my personal fork:

https://github.com/montchr/digga/actions/workflows/check.yml?query=branch%3Afeature%2Fdarwin-hosts-support

Inspired by the conversation in #16 

## TODO

- [ ] Checks succeed in each example flake
	- [x] `hmOnly`
	- [x] `groupByConfig`
	- [ ] `downstream` aka https://github.com/divnix/devos
- [ ] Rebase, squash, and clean up commits
- [ ] Revert changes made for testing purposes
	- [ ] Remove references to my Cachix store
	- [x] Remove devos fork submodule and its references
	- [ ] Remove references to my digga fork
- [x] Run builds on `macos-latest` in addition to Ubuntu
- [ ] Suggest any necessary configuration changes to https://github.com/divnix/devos in a new PR referencing this one
- [ ] Run `nix build .#mkApiReference*`
- [ ] Update nix-darwin input URL
	- [ ] If https://github.com/LnL7/nix-darwin/pull/429 is merged before this PR is ready to go, then update the input URL to point back to that repo
	- [ ] If the nix-darwin PR mentioned above is not yet merged, then consider forking nix-darwin in the divnix org
	- [ ] Otherwise, do nothing
- [ ] Test on a real Apple computer


montchr:
Thanks! Glad to hear it. And thanks for your FUP example referenced in #16, that was a big help. Though I think I lost something along the way, as the current state of this branch doesn't even use the `nixosSystems` or `darwinSystems` variables despite setting them up.

> The only main comment I have for now is that I think it would better if we could just have a Darwin example committed directly to this repo or to add a Darwin host to devos. I think we should limit the submodule based testing to the official template.

Yeah good point. Adding a Darwin example directly in the repo would also be a lot less time-consuming and complicated than trying to keep track of my devos fork in tandem with these changes. I'll update with that when I can.

src/mkFlake/fup-adapter.nix
```diff
@@ -66,7 +104,7 @@ let
       supportedSystems;
     inherit sharedOverlays;
 
-    hosts = builtins.mapAttrs (_: stripHost) config.nixos.hosts;
+    hosts = builtins.mapAttrs (_: stripHost) (config.nixos.hosts // config.darwin.hosts);
```
Pacman99, line 107: So I think here you can't give it hosts, because you already built the hosts for nixos and darwin earlier. This is where you would make use of `darwinSystems` and `nixosSystems`. I believe you can do something like this here:
```
inherit (nixosSystems) nixosConfigurations;
inherit (darwinSystems) darwinConfigurations;
```

src/mkFlake/fup-adapter.nix
```diff
@@ -58,15 +62,46 @@ let
...
+      (flake-utils-plus.lib.mergeAny
+        (stripHost config.${systemType}.hostDefaults)
+        {
+          # add `self` & `inputs` as specialargs so their libs can be used in imports
+          specialArgs = config.${systemType}.importables // { inherit (config) self inputs; };
+          modules = config.${systemType}.hostDefaults.exportedModules ++ defaultHostModules;
+        });
+  };
+
+  nixosSystem = flake-utils-plus.lib.mergeAny (diggaSystem "nixos") { };
```
blaggacao, line 86: Instead of consuming the implicit fup contract / default, we could make it explicit as in the darwin case.

That gives a slight plus in readability (by not requiring knowledge of the fup default).

src/mkFlake/options.nix
```diff
@@ -456,13 +471,30 @@ in
         The builder gets passed an attrset of all channels
       '';
     };
+    hostDefaults = mkOption {
+      type = with types; sharedHostDefaultsType;
+      default = { channelName = "nixos"; };
```
blaggacao, line 476: I think I'm ok with this. It's an implicit contract, but seems to be widespread enough.

src/modules.nix
```diff
@@ -54,8 +54,10 @@
 
       _module.args = {
         inherit hmUsers;
-        hosts = builtins.mapAttrs (_: host: host.config)
-          (removeAttrs self.nixosConfigurations [ config.networking.hostName ]);
+        hosts = builtins.mapAttrs (_: host: host.config) (
+          (removeAttrs self.nixosConfigurations [ config.networking.hostName ])
+          ++ (removeAttrs self.darwinConfigurations [ config.networking.hostName ])
+        );
```
blaggacao, line 60: This is broken, we should remove it.

It is broken since host identifiers can be reverse DNS, but `networking.hostName` will always only be the dns lable part.

Without this safeguard, if one tries to reference the current host it results in a rather obvious infinite recursion.

blaggacao (changes requested):
Looks pretty neat. I haven't been able to fully go through the module system changes on the phone.

There are just two things that I'd change as of now.

src/mkFlake/fup-adapter.nix
```diff
@@ -58,16 +62,69 @@ let
...
+      mergedDefaults = { inherit channelName system output; } //
+        (flake-utils-plus.lib.mergeAny
+          (stripHost config.hostDefaults)
+          (flake-utils-plus.lib.mergeAny
+            (stripHost config.${systemType}.hostDefaults)
+            {
+              # add `self` & `inputs` as specialargs so their libs can be used in imports
+              specialArgs = config.${systemType}.importables // { inherit (config) self inputs; };
+              modules = config.${systemType}.hostDefaults.exportedModules ++ defaultHostModules;
+            }));
```
montchr, line 81: I'm thinking that this is way more complex than it needs to be. The code is straightforward enough, but the top-level `config.hostDefaults` seems like it might not offer much benefit. Thoughts?

montchr: For example, I'd be just as happy with creating a variable in my config listing the modules shared between nixos/darwin hosts and merging that list with the system-type-specific modules list.

src/mkFlake/outputs-builder.nix
```diff
@@ -3,6 +3,7 @@
 config: channels:
 let
 
+  # TODO: too prescriptive? does not account for `config.darwin.hostDefaults`
```
blaggacao, line 6: Thanks for marking them.

Following up on my other comment that we'd only need to find a solution for those:

What would you think of if we fall back on the darwin defaults if it's `isDarwin` and otherwise on the nixos defaults?

If a user is having multiple host architectures under management and they want to re-use their hm/devshell profiles it's reasonable to expect them to take package compatibiliy across bsd/linux into account at the leaves.

blaggacao (commented):
Resolve fixturee on `nixos.hostDefaults.channelName`

src/mkFlake/outputs-builder.nix
montchr: Hey @blaggacao,

I've made some updates to address this issue. Because `isDarwin` comes from `pkgs.stdenv.isDarwin`, I've used `channels.nixpkgs.stdenv.isDarwin` to do the initial system check. Not sure if there's another way to do the `isDarwin` check without relying on _some_ version of `pkgs.stdenv`.

Relevant line: https://github.com/divnix/digga/pull/111/files#diff-b372ff5f398b3b0a0554513319e2bff34bf0cff40fee9f4f74cac0cae87ed5a6R7

How does that look?

montchr:
Made a bit more progress here over the past day. Currently running into the following issue in both the `groupByConfig` and the `nix-darwin` examples:

```
trace: { config = <CODE>; }
error: module function has no (required) _file argument key: .
```

Not quite sure what the exact cause is, I'll have to dig in a little further. Any pointers would be helpful 🙂 

Full trace:

```
❯ nix flake check --show-trace
warning: Git tree '/Users/montchr/Developer/contrib/divnix/digga' is dirty
warning: creating lock file '/Users/montchr/Developer/contrib/divnix/digga/examples/groupByConfig/flake.lock'
warning: Git tree '/Users/montchr/Developer/contrib/divnix/digga' is dirty
warning: unknown flake output 'pkgs'
warning: unknown flake output 'homeConfigurationsPortable'
trace: { config = <CODE>; }
error: module function has no (required) _file argument key: .


       … while evaluating anonymous lambda

       at /nix/store/d3vlpgh54ackhc03fdcn9pqh5q3cgkcp-source/lib/exportModules.nix:31:8:

           30|     genAttrs'
           31|       (arg:
             |        ^
           32|

       … from call site

       … while evaluating 'genAttrs''

       at /nix/store/d3vlpgh54ackhc03fdcn9pqh5q3cgkcp-source/lib/internal-functions.nix:13:21:

           12|   */
           13|   genAttrs' = func: values: builtins.listToAttrs (map func values);
             |                     ^
           14|

       … from call site

       at /nix/store/d3vlpgh54ackhc03fdcn9pqh5q3cgkcp-source/lib/exportModules.nix:30:5:

           29|
           30|     genAttrs'
             |     ^
           31|       (arg:

       … while evaluating 'exportModules'

       at /nix/store/d3vlpgh54ackhc03fdcn9pqh5q3cgkcp-source/lib/exportModules.nix:10:19:

            9|     ;
           10|   exportModules = args:
             |                   ^
           11|     /**

       … from call site

       at /nix/store/25wbm1hxrnwqkv2harng487wshb8qdy9-source/src/mkFlake/fup-adapter.nix:151:20:

          150|
          151|     nixosModules = flake-utils-plus.lib.exportModules (diggaSystemModules "nixos");
             |                    ^
          152|

       … while checking flake output 'nixosModules'

       at /nix/store/25wbm1hxrnwqkv2harng487wshb8qdy9-source/src/mkFlake/fup-adapter.nix:151:5:

          150|
          151|     nixosModules = flake-utils-plus.lib.exportModules (diggaSystemModules "nixos");
             |     ^
          152|
2.19s user 14.06s system 57% cpu 28.385s total
```

liaowang11:
I'm trying to get this pr to work with devos template, according to the doc, I added this to `flake.nix`: 
```nix
        darwin = {
          hostDefaults = {
            system = "x86_64-darwin";
            channelName = "nixos";
            imports = [];
            modules = [
              { lib.our = self.lib; }
              digga.nixosModules.nixConfig
              home.nixosModules.home-manager
              agenix.nixosModules.age
              bud.nixosModules.bud
            ];
          };
          hosts = {
              /* set host specific properties here */
              "mac-host1" = {
                system = "x86_64-darwin";
                modules = [
                  { home-manager.useUserPackages = false; }
                  ./darwin/darwin-configuration.nix
                ];
              };
          };

          importables = rec {
            profiles = digga.lib.rakeLeaves ./profiles // {
              users = digga.lib.rakeLeaves ./users;
            };
            suites = with profiles; rec {
              mac = [  ];
            };
          };
        };
```

But when I build it with `nix build --show-trace ".#darwinConfigurations.mac-host1.system"`, there is a weird error which is hard to debug, really don't know what's the problem as I'm not so familiar with digga code. Would you be so kind to help?
```
error: The option `lib' does not exist. Definition values:
       - In `<unknown-file>'

       … while evaluating the attribute 'config.system.build.toplevel'
```

montchr:
@liaowang11 indeed – nothing works here yet. I'm sorry that I haven't made that clear in my updates. The "option `lib' does not exist" error is a new one – I pushed a change the other day which introduced it – and I'm hopeful that I'll be able to figure that one out since it isn't as cryptic as some of the other errors I've seen (more on that below).

---

To all curious observers: **this draft pull request is not yet functional**.

I have not yet been able to get `nix develop` to succeed in any of the example flakes – it fails with an error every time, despite my efforts to make progress.

I would appreciate any help – any insight into the cause of the error messages in CI would move this PR along.

You can see the current state of this PR's builds on my personal fork:

https://github.com/montchr/digga/actions/workflows/check.yml?query=branch%3Afeature%2Fdarwin-hosts-support

I haven't posted an update in a while because, to be honest, I have run into build errors every step of the way and it's gotten quite frustrating. Every so often I try my hand at making changes with fresh eyes, but usually the errors are cryptic. In general, I'm working in the dark.

---

This has been the most cryptic and persistent error I've run into:

https://github.com/montchr/digga/runs/4830689744?check_suite_focus=true

```
error: value is null while a string was expected

       … while evaluating 'seive'

       at /nix/store/nvb1mhmwdqhwm0bzv6dd890hgrgaf5zf-source/src/mkFlake/outputs-builder.nix:76:22:

           75|         let
           76|           seive = _: v: v.system == system; # only test for the appropriate system
             |                      ^
           77|           collectActivationPackages = n: v: { name = "user-" + n; value = v.activationPackage; };

       … from call site

       at /nix/store/bxapqbai4fnwm7297y02m9gjhz9w959n-source/lib/attrsets.nix:128:62:

          127|   filterAttrs = pred: set:
          128|     listToAttrs (concatMap (name: let v = set.${name}; in if pred name v then [(nameValuePair name v)] else []) (attrNames set));
             |                                                              ^
          129|

       … while evaluating anonymous lambda

       at /nix/store/bxapqbai4fnwm7297y02m9gjhz9w959n-source/lib/attrsets.nix:128:29:

          127|   filterAttrs = pred: set:
          128|     listToAttrs (concatMap (name: let v = set.${name}; in if pred name v then [(nameValuePair name v)] else []) (attrNames set));
             |                             ^
          129|

       … from call site

       at /nix/store/bxapqbai4fnwm7297y02m9gjhz9w959n-source/lib/attrsets.nix:128:18:

          127|   filterAttrs = pred: set:
          128|     listToAttrs (concatMap (name: let v = set.${name}; in if pred name v then [(nameValuePair name v)] else []) (attrNames set));
             |                  ^
          129|

       … while evaluating 'filterAttrs'

       at /nix/store/bxapqbai4fnwm7297y02m9gjhz9w959n-source/lib/attrsets.nix:127:23:

          126|   */
          127|   filterAttrs = pred: set:
             |                       ^
          128|     listToAttrs (concatMap (name: let v = set.${name}; in if pred name v then [(nameValuePair name v)] else []) (attrNames set));

       … from call site

       at /nix/store/nvb1mhmwdqhwm0bzv6dd890hgrgaf5zf-source/src/mkFlake/outputs-builder.nix:79:9:

           78|         in
           79|         lib.filterAttrs seive (lib.mapAttrs' collectActivationPackages config.self.homeConfigurations)
             |         ^
           80|       else { }

       … while evaluating the attribute 'checks'

       at /nix/store/nvb1mhmwdqhwm0bzv6dd890hgrgaf5zf-source/src/mkFlake/outputs-builder.nix:68:3:

           67|
           68|   checks =
             |   ^
           69|     (

       … while evaluating anonymous lambda

       at /nix/store/zzp9cqzfqmsqxkxbz3k17q5kp5n48xsv-source/flake.nix:34:20:

           33|           lhs // mapAttrs
           34|             (name: value:
             |                    ^
           35|               if isAttrs value then lhs.${name} or { } // value

       … from call site
git+file:///home/runner/work/digga/digga/examples/downstream
```

I think I've seen this in multiple example check tasks (e.g. `downstream`, `hmOnly`, `groupByConfig`) – the error above comes from a `downstream` check. I've disabled `downstream` checks for the time being to keep things simple by only testing the examples defined in the digga repo. Once I'm able to run `nix develop` via CI on `hmOnly` and `groupByConfig`, I'll then re-enable `downstream` checks.

src/mkFlake/outputs-builder.nix
```diff
@@ -105,6 +109,7 @@ in
     (
       # for config.self.nixosConfigurations if present & non-empty
       if (
+        (hostType == "linux") &&
```
liaowang11, line 112: hosType == "nixos" ?

src/mkFlake/outputs-builder.nix
```diff
@@ -3,7 +3,11 @@
 config: channels:
 let
 
-  pkgs = channels.${config.nixos.hostDefaults.channelName};
+  defaultPkgs = channels.${config.nixos.hostDefaults.channelName};
+  hostType = if defaultPkgs.stdenv.isDarwin then "darwin" else "nixos";
+  hostTypeDefaultChannel = config.${hostType}.hostDefaults.channelName;
+
+  pkgs = channels.${hostTypeDefaultChannel};
   system = pkgs.system;
```
liaowang11, line 11: I'm not quite sure if this should be `system=defaultPkgs.system`

liaowang11:
I did some digging for a couple of hours, didn't find a lot of useful information. But, the error `﻿﻿value is null while a string was expected` seems to be related nix. I noticed this when trying to use [the debugger pr to nix](https://github.com/NixOS/nix/pull/5416/files). After upgrading to the latest nix version and manually run `nix flake show` in the `examples/hmOnly`, it works.
```
❯ nix --version
nix (Nix) 2.6.0pre20211217_6e6e998
```
```
❯ nix flake show
├───checks
│   ├───aarch64-darwin
│   │   └───user-testuser: derivation 'home-manager-generation'
│   ├───aarch64-linux
│   │   └───user-testuser: derivation 'home-manager-generation'
│   ├───i686-linux
│   │   └───user-testuser: derivation 'home-manager-generation'
│   ├───x86_64-darwin
│   │   └───user-testuser: derivation 'home-manager-generation'
│   └───x86_64-linux
│       └───user-testuser: derivation 'home-manager-generation'
├───devShell
│   ├───aarch64-darwin: development environment 'nixos'
│   ├───aarch64-linux: development environment 'nixos'
│   ├───i686-linux: development environment 'nixos'
│   ├───x86_64-darwin: development environment 'nixos'
│   └───x86_64-linux: development environment 'nixos'
├───devshellModules: unknown
├───homeConfigurationsPortable: unknown
├───homeModules: unknown
├───nixosModules
├───overlays
├───packages
│   ├───aarch64-darwin
│   ├───aarch64-linux
│   ├───i686-linux
│   ├───x86_64-darwin
│   └───x86_64-linux
└───pkgs: unknown

```

---
So I just run `nix develop --command -- check-hmOnly`, it still failed with this error. Note: I updated the nixpkgs of digga to latest, so the nix version is the latest.
```
❯ nix develop --command -- check-hmOnly
nix version:
nix (Nix) 2.6.0pre20211217_6e6e998
├───checks
│   ├───aarch64-darwin
error: value is null while a string was expected
(use '--show-trace' to show detailed location information)
rm 'examples/hmOnly/flake.lock'
exit 1: command [/nix/store/28gq8irvyxwm023a4zhmf6n2il88zmym-nix-2.6pre20211217_6e6e998/bin/nix flake show "$@"] failed
```

src/mkFlake/outputs-builder.nix
montchr: Good catch! You're right, thanks.

montchr:
I've just force-pushed a pretty significant rewrite to this branch, and I have a sense that I've made some good progress here. Builds are still failing, but the types of errors I'm seeing now appear much more promising – the latest error I'm seeing _on both `ubuntu` and `macos-latest` is referring to the darwin example host `Darwinia`.

https://github.com/montchr/digga/runs/4914183517?check_suite_focus=true

Other notes:

- I wanted to take as much "magic" as possible out of the new changes to `fup-adapter.nix` to make it easier to debug. So there may be some opportunity for introducing some DRY functions in that file down the line, but for now I'd like to keep things less complex by introducing some WET code.
- I realized that the toplevel location of the "option `lib' does not exist" error resulted from differences between NixOS (which declares a top-level `lib` option) and nix-darwin. Sure enough, nix-darwin does not currently declare a top-level `lib` option. To that end, I've opened https://github.com/LnL7/nix-darwin/pull/429. Because this one is kind of a show-stopper, I've hard-coded the `darwin` input's URL to the branch on my fork which the nix-darwin PR is based upon. Once that's merged, we can switch back to the main URL. If that doesn't happen soon enough, then I'd suggest forking nix-darwin to the divnix org. I've added these to the checklist in the PR description.
- I think the introduction of an intermediary flake (i.e. the approach suggested in #16) made it much harder to trace errors that were occurring during the build of the intermediary flake. Maybe it can be changed back later, but if it's possible to accomplish #16 without double-flaking it, in my opinion that would be preferable.

montchr:
I'm very happy to report what looks like the first set of passing builds since I started working on this:

https://github.com/montchr/digga/runs/4917601908?check_suite_focus=true

Next up is adding some nix-darwin-specific configuration and digging in to verify a little more closely. With these passing builds, I'm hopeful that I'll be able to test drive a digga-powered darwin system configuration pretty soon!

src/mkFlake/fup-adapter.nix
```diff
@@ -51,20 +51,71 @@ let
...
     inherit (config)
       channelsConfig
       supportedSystems;
     inherit self inputs sharedOverlays;
 
-    hosts = builtins.mapAttrs (_: stripHost) config.nixos.hosts;
+    hosts = builtins.mapAttrs (_: stripHost) (
+      flake-utils-plus.lib.mergeAny
+        nixosHosts
+        darwinHosts
```
Pacman99, line 117:
```suggestion
      flake-utils-plus.lib.mergeAny
        nixosHosts
        darwinHosts
```
While I realize that this makes the code easier, the issue with this is that darwin and nixos hosts can't have the same name even though this is actually possible. This creates an artificial restriction on host names. If we do decide to do this, there needs to be an error triggered early when a user passes a host to both `nixos.hosts` and `darwin.hosts` so they know that only one will be created.

src/mkFlake/fup-adapter.nix
```diff
@@ -75,14 +126,10 @@ let
       )
       config.channels;
 
-    hostDefaults = flake-utils-plus.lib.mergeAny (stripHost config.nixos.hostDefaults) {
```
Pacman99, line 78: I was thinking it would be nice if `hostDefaults` could be a function when passed to fup, so that you could dynamically set defaults.
```
hostDefaults = hostAttrs: { system = if hostAttrs.output = "darwinConfigurations" then "x86_64-darwin" else "x86_64-linux"; }
```

src/mkFlake/fup-adapter.nix
```diff
@@ -51,20 +51,71 @@ let
...
+
+      # add `self` & `inputs` as specialArgs so their libs can be used in imports
+      specialArgs = config.nixos.importables // { inherit (config) self inputs; };
+
+      modules = config.nixos.hostDefaults.exportedModules ++ defaultHostModules ++ [
+        ({ config, pkgs, self, ... }: {
+          users.mutableUsers = lib.mkDefault false;
+          hardware.enableRedistributableFirmware = lib.mkDefault true;
+          system.configurationRevision = lib.mkIf (self ? rev) self.rev;
+        })
```
Pacman99, line 76: I think it would be better if this was put in `src/modules.nix` as `nixosDefaults`

Pacman99 (commented):
Everything is looking really good! Sorry I couldn't help with debugging more. It seems that things are coming along well

src/mkFlake/fup-adapter.nix
montchr: Good point, though I assumed that would be an edge case. There is a unique-name restriction amongst `nixos.hosts` already, and I would think that allowing identical hostnames across `{nixos,darwin}.hosts` would be difficult to manage and could lead to weird issues down the road. So I'm thinking that throwing an error for any identical hostnames would be the better option here.

Pacman99:
So due to all the confusion with #divnix/digga#431, I went ahead and rebased your commits on the new digga main for you. You can find it in divnix/digga/darwin-support. Feel free to add that to your devos fork(and rename that to digga) and re-create the PR. I was also thinking we could give you commit access to digga, so you can work on the darwin-support branch directly.

montchr:
Closing in favor of a new PR at https://github.com/divnix/digga/pull/438/

PR #114, opened 2021-11-26
Follow release-21.11, use nixStable, nix 2.4 with flakes support

teutat3s:
- Update nixos input to new stable branch `release-21.11`
- [nix 2.4 with flakes support](https://discourse.nixos.org/t/nix-2-4-released/15822) has landed in 21.11, let's use it

Not sure what to do about the circular dependency in the `Checks` GH action.

I guess we could merge this, then update https://github.com/divnix/devos with the new commit in its `flake.lock`, then bump the git submodule here in digga.

Please tell me if you'd like to leave patched-nix commented out as an example or would rather have it deleted.

Looking forward for feedback to improve this PR.


teutat3s:
For the curious I have prepared a branch to create a follow-up PR for devos:

https://github.com/divnix/devos/compare/main...pub-solar:bump/release-21.11

overlays/default.nix
```diff
@@ -1,3 +1,4 @@
 { inputs }: {
-  patchedNix = import ./nix-patched.nix { inherit inputs; };
```
blaggacao, line 2: I think you can remove this entirely.
For conciseness sale, we don't really need (to keep) code examples here in the lob.

examples/hmOnly/flake.nix
```diff
@@ -3,11 +3,11 @@
 
   inputs =
     {
-      nixos.url = "github:nixos/nixpkgs/release-21.05";
-      digga.url = "path:../../";
+      nixos.url = "github:nixos/nixpkgs/release-21.11";
+      digga.url = "github:divnix/digga";
```
blaggacao, line 7: Have you checked how this affects the obscure `sed` logic in the devshell?

examples/groupByConfig/flake.nix
```diff
@@ -3,9 +3,9 @@
 
   inputs =
     {
-      nixos.url = "github:nixos/nixpkgs/release-21.05";
+      nixos.url = "github:nixos/nixpkgs/release-21.11";
       digga = {
-        url = "path:../../";
+        url = "github:divnix/digga";
```
blaggacao, line 8: ditto

blaggacao (commented):
Thanks for bumping! Please find my comments attached.

examples/hmOnly/flake.nix
blaggacao: I juat checked, I think `digga_fixture` will still do it's thing and try to avoid fetching remotely from github.

The problem was that if it tries to fetch, it doesn't fetch the latest commit if that commit is only locally available, yet.

teutat3s: I'll try to double check that, but without this change, I saw this error:
```
[digga]$ check-groupByConfig --show-trace
warning: unknown setting 'default-flake'
error: found circular import of flake 'path:./'

       … while updating the flake input 'digga/digga'

       … while updating the flake input 'digga'

       … while updating the lock file of flake 'git+file:///home/teutat3s/CodeRoom/github.com/divnix/digga?dir=examples%2fgroupByConfig&ref=main&rev=f751001f921fde31877295ea0155426f4f18f832'
exit 1: command [/nix/store/ag01yxbp4z0bksh7qc67xc31rgk91yaa-nix-2.4/bin/nix flake lock --update-input digga "$@"] failed
```

overlays/default.nix
teutat3s: Agreed. Done.

examples/hmOnly/flake.nix
blaggacao: Oh shoot, I almost suspected this. That means the digga fixture must implement pinni g to the current rev with somestimg like `git ref blabla` (don't remember the command).

Else a local check will happily fool anyone by taking whatever is "latest" on remote.

If things are not pushed, people think "all good" while in reality the check has been run on a different commit.

It is also a problem, when we have a fork, since the url will not be able to capture a fork's commit the way this is set up now.

blaggacao:
You can also remove the advisory in the readme (and recreate docs, which consume the readme). Finally :smile:

examples/hmOnly/flake.nix
blaggacao: @CitadelCore I ping you since I beleive, you had implemented some of these safeguards when you implemented the follows patch. Can you share your thoughts on this problem?

CitadelCore: That doesn't look like any safeguard I added, that's an existing error from Nix. It looks like there may need to be some special handling implemented in order for it to handle parent relative paths properly.

nrdxp:
We won't be able to use `nixStable` and 21.11 together once this is merged, and it likely will be today or tomorrow.
NixOS/nixpkgs#147511

blaggacao (approved):
I'm approving this and will apply the `nixStable` -> `nixUnstable` revert directly on master since I can't modify this PR.

blaggacao:
bors try

blaggacao:
Merged with passing `check-all`.

PR #117, opened 2021-12-08
nix: remove ca-references feature flag

nrdxp:
This experimental feature has been removed from Nix in unstable versions recently, and with this set, it now pops up an unsupressable warning message. Not so bad on its own, but when using <kbd>Tab</kbd> completion it corrupts the output.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -92,11 +92,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1630859749,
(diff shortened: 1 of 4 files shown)


blaggacao (approved):

Pacman99 (approved):

nrdxp:
something is up with CI, should I just merge?

Pacman99:
I think the CI error is in devos, we need to merge the update PR

PR #121, opened 2022-01-14
Fix default input detection

Pacman99:
This fixes divnix/devos#413 and the general infinite recursion issues that are happening.
Also switch back to upstream flake-utils-plus, the divnix patches. I don't believe the divnix patches actually affect anything.

### DIFF
--- a/doc/api-reference-channels.md
+++ b/doc/api-reference-channels.md
@@ -48,7 +48,7 @@ nix flake
 
 *_Default_*
 ```
-"config.self.inputs.<name>"
+"self.inputs.<name>"
 ```
 
 
--- a/flake.lock
+++ b/flake.lock
@@ -88,15 +88,15 @@
         "flake-utils": "flake-utils"
       },
       "locked": {
-        "lastModified": 1638994888,
-        "narHash": "sha256-iz/ynGNZlvqKCOnFrEKqGA+BVKGQMG+g2JT+e3OOLN8=",
-        "owner": "divnix",
+        "lastModified": 1639385028,
+        "narHash": "sha256-oqorKz3mwf7UuDJwlbCEYCB2LfcWLL0DkeCWhRIL820=",
+        "owner": "gytis-ivaskevicius",
         "repo": "flake-utils-plus",
-        "rev": "b4f9f517574cb7bd6ee3f19c72c19634c9f536e1",
+        "rev": "be1be083af014720c14f3b574f57b6173b4915d0",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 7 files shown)


jackmac92:
awesome! let me know if I can help by testing 
could just use a tip on where the branchname should go. Can I just append it [here](https://github.com/divnix/devos/blob/283f6b255e4371c13c0f4dd9320f56613b49cb86/flake.nix#L13)?

Pacman99:
Yes! It would be great if you could run some tests with this branch and ensure the infinite recursion errors get fixed.
Just set the digga input line to:
`digga.url = "github:Pacman99/digga/default-input-detection`.

jackmac92:
awesome (really appreciate your help 🙇 )

gave it a shot I am seeing a new inf recursion error.

Just for context, w/in the VM I create with virt-manager I see this
```
root in /iso/devos
❯ cat flake.lock  | grep -C 4 Pacman
      },
      "locked": {
        "lastModified": 1642140093,
        "narHash": "sha256-4PZyuXtYXP8F4oiJuWQjQz3kObQaxJDTOQ6q3LtUm+k=",
        "owner": "Pacman99",
        "repo": "digga",
        "rev": "ee307797d398fbfa2dab778b1dfa09c97daa4f3a",
        "type": "github"
      },
      "original": {
        "owner": "Pacman99",
        "ref": "default-input-detection",
        "repo": "digga",
        "type": "github"
      }
```
(trying to confirm I'm using the correct branch/setup)

this is the error `nix-shell` shows (full trace attached as a file)

```
error: infinite recursion encountered

       at /nix/store/bxapqbai4fnwm7297y02m9gjhz9w959n-source/lib/trivial.nix:14:8:

           13|     # The value to return
           14|     x: x;
             |        ^
           15|
(use '--show-trace' to show detailed location information)
```
[nix-shell trace](https://github.com/divnix/digga/files/7875781/nix-shell.trace.txt)

blaggacao (approved):

Pacman99:
Ok I think I've got all of it this time. So it seems that `inherit (config) self` works with `nix develop`, but not `nix-shell`. We need to make sure we explicitly use `inherit self inputs` from now on, so we don't have underlying infinite recursion errors.

ModdedGamers:
Could you update your branch? This branch has the problem fixed by ea4af3204c, and so it fails. If you rebase, I'd love to test this.

Pacman99:
> Could you update your branch?

done, thanks!

blaggacao (approved):

PR #115, opened 2021-11-30
cleanup dar

blaggacao:
- rm: finally deprecated code
- imp: add .envrc for flakes / non flakes variants
- fix: update & fix docs
- treefmt: impl & apply
- rm: ca-references flag
- imp: remove (all) most follow instruction from digga

PR #122, opened 2022-01-14
flake: remove anti-corruption layer and use nixlib

Pacman99:
This means the only remaining part of digga lib that uses nixpkgs is deploy, which can hopefully be removed with use of the deploy overlay instead of `deploy.lib`.

A small piece of #115 focusing on the follows changes

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -21,9 +21,7 @@
         "nixpkgs": [
           "latest"
         ],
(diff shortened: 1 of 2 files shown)


Pacman99:
bors try

blaggacao (approved):

PR #124, opened 2022-01-22
remove vm of nixos-generators.nixosModules on 22.05

### LINKED ISSUE #123: `system.build.vm` from nixpkgs and nixos-generators conflicts
On the unstable channel, `system.build.vm` can not be evaluated, so all vm tests are broken.

In the `customBuilds` module of digga, all builds from nixos-generators are added into `system.build`, including `vm`.
https://github.com/divnix/digga/blob/fca1df3dcdb78531772baf801b79e01489280864/src/modules.nix#L16-L26

Recently https://github.com/NixOS/nixpkgs/pull/151082 was merged into nixpkgs. Now nixpkgs also add `system.build.vm` to configurations, causing this issue.
https://github.com/NixOS/nixpkgs/blob/3cbb02087b4c2dfc3be664d7152cec51765e53ef/nixos/modules/virtualisation/build-vm.nix#L49-L52
```nix
    system.build = {
      vm = lib.mkDefault config.virtualisation.vmVariant.system.build.vm;
      vmWithBootLoader = lib.mkDefault config.virtualisation.vmVariantWithBootLoader.system.build.vm;
    };
```

Nixpkgs actually use `lib.mkDefault`, but the issue still happens, very wired. After apply the possible solution below, the issue disppears. So this might also be a nixos module issue or nix issue.

## Expected Behavior

`system.build.vm` can be evaluated and built.

## Current Behavior

My nix is `github:nixos/nix/bc443511eb65420b51d10708e25427fe50de37a8`. When I run `nix build .#nixosConfigurations.<someHost>.system.build.vm`, nix just stuck and eats all of my memory (24GB)... I have no idea about why this happens.

Also reproducible using `github:nixos/nixpkgs/d5dae6569ea9952f1ae4e727946d93a71c507821#nixUnstable`.

## Possible Solution

Move all custom builds to another location, for example `system.build.customBuilds`.
```diff
diff --git a/src/modules.nix b/src/modules.nix
index 50c78be..9fcd5f9 100644
--- a/src/modules.nix
+++ b/src/modules.nix
@@ -13,7 +13,7 @@
           # so try to pass that to eval if possible.
           specialArgs = args.specialArgs or { };
         };
-      system.build =
+      system.build.customBuilds =
         let
           builds = lib.mapAttrs
             (format: module:
```

## Steps to Reproduce

1. Download the minimal `flake.nix` and `flake.lock` from my gist https://gist.github.com/linyinfeng/50a6569f57694c189ac872e868a8df85.
2. Run `nix build .#nixosConfigurations.test.system.build.vm`.
3. Monitor memory usage. Kill nix before the machine freezes up.

## Context

Use vm tests to check my configuration.

## Your Environment

1. nix `bc443511eb65420b51d10708e25427fe50de37a8`
2. nixpkgs `d5dae6569ea9952f1ae4e727946d93a71c507821`

GTrunSec:
try to fix #123

### DIFF
--- a/src/modules.nix
+++ b/src/modules.nix
@@ -17,10 +17,13 @@
         let
           builds = lib.mapAttrs
             (format: module:
-              let build = config.lib.digga.mkBuild module; in
+              let build = config.lib.digga.mkBuild module;
+              in
               build // build.config.system.build.${build.config.formatAttr}
             )
-            nixos-generators.nixosModules;
+            (if (lib.versionAtLeast config.system.stateVersion "22.05") then
+              builtins.removeAttrs nixos-generators.nixosModules [ "vm" ]
+            else nixos-generators);
         in
         # ensure these builds can be overriden by other modules
         lib.mkBefore builds;
@@ -63,4 +66,3 @@
       system.configurationRevision = lib.mkIf (self ? rev) self.rev;
     };
 }
-


GTrunSec:
@Pacman99  ready to review. thanks.

Pacman99 (approved):
This LGTM! What do you think about putting all the custom builds under `system.build.customBuilds` as was suggested in #123? I think it might be the right move for long term, but also having everything in `system.build` has a nice elegance/simplicity to it.

GTrunSec:
Yes, It could be a solution for the long term. But at this moment I don't think it's necessary to move them all in `customBuild`. I prefer to set the `nixos-generators`  as the default of `system.build.<args>`. Also, From #123, the issue only conflicted with the vm.module.

src/modules.nix
```diff
@@ -17,10 +17,13 @@
...
           builds = lib.mapAttrs
             (format: module:
-              let build = config.lib.digga.mkBuild module; in
+              let build = config.lib.digga.mkBuild module;
+              in
               build // build.config.system.build.${build.config.formatAttr}
             )
-            nixos-generators.nixosModules;
+            (if (lib.versionAtLeast config.system.stateVersion "22.05") then
+              lib.filterAttrs (n: v: n != "vm") nixos-generators.nixosModules
```
Pacman99, line 25: Actually could we use `builtins.removeAttrs` for this instead?

GTrunSec: Sure, thanks.

Pacman99 (approved):
Cool LGTM, thanks! And yeah I agree with your reasoning, for now this just seems to be an exception
bors r+

PR #126, opened 2022-01-30
Cleanup how polyfill works for digga-use outputs

Pacman99:
I feel like the polyfill code could be done a lot simpler without us having to maintain a local copy for flake-compat. And I never liked the function `ufrContract` because I would have to lookup what it was doing every time it was used. So I think its better for that logic to be done in the `flake.nix` and to use `genAttrs` that makes this much simpler to implement.


blaggacao (commented):
Good idea!

checks/default.nix
```diff
@@ -1,12 +1,11 @@
...
+, inputs ? (import ./.).inputs
 }:
 let
 
-  nixpkgs = inputs.nixpkgs;
-  digga = inputs.digga;
+  inherit (inputs) digga nixpkgs;
   lib = nixpkgs.lib // digga.lib;
-  pkgs = import nixpkgs { inherit system; config = { }; overlays = [ ]; };
+  pkgs = nixpkgs.legacyPackages.${system};
```
Pacman99, line 8: We only support systems that nixpkgs supports, so we may as well just use the flake output. Also I think this will be even better for eval caching.

Pacman99:
What was `jobs/flake.nix` for? I can't tell if needs to stay, and if it does I think it needs to be updated with this method.

Pacman99:
bors try

Pacman99:
bors try

Pacman99:
I know the unofficial flakes roadmap restricts arguments to `system` and `input`, but I wonder if it makes more sense to do `self` and `system`. That would make polyfilling much simpler to implement. And it would be flow better for readers. Code would access the flakes inputs through `self.inputs` instead of accessing itself with `inputs.digga`.

